PR #27, opened 2023-09-01
Optional plotting dependency

McHaillet:
The estimate_roc script requires matplotlib and seaborn, so adding them as optional dependencies is the best way to do it.


sroet (approved):
LGTM, feel free to merge

Issue #25, opened 2023-08-31
Error running the test


hamid13r:
Hi,

I just installed pytom_tm, we use micromamba so I had to install cupy from conda-forge `conda install -c conda-forge cupy`. I still get an error when running the tests:

```
> ...Progress job_0 on device 0:
> 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:01<00:00, 217.84it/s]
> .WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> Progress job_0 on device 0:
> 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:01<00:00, 243.14it/s]
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> DEBUG:root:Next fast fft shape: (100, 45, 60)
> Progress job_00 on device 0:
> 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:01<00:00, 237.16it/s]
> DEBUG:root:Next fast fft shape: (100, 48, 60)
> Progress job_01 on device 0:
> 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:01<00:00, 249.47it/s]
> DEBUG:root:Next fast fft shape: (100, 45, 60)
> Progress job_02 on device 0:
> 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:01<00:00, 249.58it/s]
> FWARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> .WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
> WARNING: Provided voxel size does not match voxel size annotated in tomogram/template mrc.
[...]

PR #31, opened 2023-09-01
added extra template file that is annotated with wrong voxel size

McHaillet:
Voxel size warning was printed a lot during unittest after I changed the template to have the wrong voxel size for an error assertion test. I know added another template with the wrong voxel size for that specific assertion test. Now the unittest does not clutter the terminal with voxel size warnings (which seems quite alarming).

As shown in issue #25.


sroet (approved):
LGTM, feel free to merge

Issue #26, opened 2023-08-31
Voxel size not identical error


hamid13r:
Hi,

I get this error "Input tomogram voxel spacing is not identical in each dimension!" when I use tomograms generated in WARP. But the template matching runs with IMOD generated tomos. 

Is there a way of forcing the pixel size?

pytom_match_template.py  -t 50S_ref.mrc -m 50S_mask.mrc -v tomo/JTL013_1_L04_ts_003.mrc_13.28Apx.mrc -a tomo/JTL014_1_L14_ts_003.mrc.tlt --angular-search 7.00 --search-x 10 500 --search-y 10 710 --search-z 10 210  --voxel-size-angstrom  13.28 --low-pass 40 --high-pass 500 -g 0 --log info

> Traceback (most recent call last):
>   File "/home/janlevinson/micromamba/envs/pytom_tm/bin/pytom_match_template.py", line 92, in <module>
>     main()
>   File "/home/janlevinson/micromamba/envs/pytom_tm/bin/pytom_match_template.py", line 62, in main
>     job = TMJob(
>           ^^^^^^
>   File "/home/janlevinson/micromamba/envs/pytom_tm/lib/python3.11/site-packages/pytom_tm/tmjob.py", line 83, in __init__
>     meta_data_tomo = read_mrc_meta_data(self.tomogram)
>                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/janlevinson/micromamba/envs/pytom_tm/lib/python3.11/site-packages/pytom_tm/io.py", line 82, in read_mrc_meta_data
>     raise ValueError('Input tomogram voxel spacing is not identical in each dimension!')
> ValueError: Input tomogram voxel spacing is not identical in each dimension!

I also get a warning when specifying `--input-voxel-size-angstrom` for `pytom_create_template.py` saying

`WARNING:root:Provided voxel size does not match voxel size annotated in input map.`
And they might be related?

Thank you for your help in advance,
Hamid

hamid13r:
Update:
I used 
`alterheader tomo/JTL013_1_L04_ts_003.mrc_13.28Apx.mrc -cel 512,722,226`
which changes the pixel size to 1,1,1 in the header. This overcomes the previous error and I only get a warning. 
```

> 
[...]

PR #33, opened 2023-09-01
patch for handling input which has unequal spacing

McHaillet:
Previous PR created some merge conflicts because main and this both changes test_tmjob.py. So this should do it.

- Function should mention volume instead of tomogram as it not only applies for tomograms.
- added error class for unequal spacing
- user is notified whether template or tomogram has unequal spacing
-  added unittest for assertion

See issue #26

### DIFF
--- a/src/pytom_tm/io.py
+++ b/src/pytom_tm/io.py
@@ -69,9 +69,7 @@ class ParseTiltAngles(argparse.Action):
 
 
 class UnequalSpacingError(Exception):
-    def __init__(self, message):
-        # Call the base class constructor with the parameters it needs
-        super().__init__(message)
+    pass
 
 
 def write_angle_list(data: npt.NDArray[float], file_name: pathlib.Path, order: tuple[int, int, int] = (0, 2, 1)):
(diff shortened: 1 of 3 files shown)


src/pytom_tm/io.py
```diff
@@ -68,6 +68,12 @@ def __call__(self, parser, namespace, values: Union[list[str, str], str], option
             parser.error("{0} can only take one or two arguments".format(option_string))
 
 
+class UnequalSpacingError(Exception):
+    def __init__(self, message):
+        # Call the base class constructor with the parameters it needs
+        super().__init__(message)
```
sroet, line 74: is this needed? could you not just replace this with:
```suggestion
    pass
```

sroet (changes requested):
1 proposed change, LGTM otherwise (please make sure the test pass after the change if you decide to implement it)

src/pytom_tm/io.py
McHaillet: you are right, that also works

sroet (approved):
LGTM, feel free to merge

PR #34, opened 2023-09-01
added non-spherical mask option to pytom_match_template.py

### LINKED ISSUE #21: non spherical mask option should be available in pytom_match_template.py
A non spherical mask flag should be added in the script: https://github.com/SBC-Utrecht/pytom-template-matching-gpu/blob/main/src/bin/pytom_match_template.py. It is already enabled in the code but forgot to yet add it as a flag to the script.

McHaillet:
Forgot to add this option to the script that was already implemented in the code.

Closes #21.

### DIFF
--- a/src/bin/pytom_match_template.py
+++ b/src/bin/pytom_match_template.py
@@ -70,7 +70,7 @@ def main():
         args.mask,
         args.destination,
         angle_increment=args.angular_search,
-        mask_is_spherical=True if args.non_spherical_mask is None else args.non_spherical_mask,
+        mask_is_spherical=True if args.non_spherical_mask is None else (not args.non_spherical_mask),
         tilt_angles=args.tilt_angles,
         tilt_weighting=args.per_tilt_weighting,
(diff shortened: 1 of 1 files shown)


src/bin/pytom_match_template.py
```diff
@@ -67,7 +70,7 @@ def main():
         args.mask,
         args.destination,
         angle_increment=args.angular_search,
-        mask_is_spherical=True,
+        mask_is_spherical=True if args.non_spherical_mask is None else args.non_spherical_mask,
```
sroet, line 73: shouldn't this be
```suggestion
        mask_is_spherical=True if args.non_spherical_mask is None else (not args.non_spherical_mask),
```

sroet (changes requested):
I think the logic is wrong, LGTM otherwise

McHaillet:
> I think the logic is wrong, LGTM otherwise

indeed :+1:

sroet (approved):
LGTM, feel free to merge

PR #38, opened 2023-09-06
default mrc metadata reading now is permissive

### LINKED ISSUE #35: issues with reading mrc files
Hey!

a) thank you for this update to pytom! I'm really excited to try GPU-accelerated TM!

b) I do not know why, but mrcfile complains when trying to read the header of our mrc files.

They work fine in any other program though ...

From the code, it appears that you use it to also get the size of the tomograms and hence I guess it cannot be avoided.

To fix this problem, the permissive statement is needed and hence I changed pytom_tm/io.py, line 79 to:

with mrcfile.mmap(file_name, **permissive=True**) as mrc:

which appears to fix the issue.

Best,

Philipp.

McHaillet:
Small fix to read mrc metadata in permissive mode by default.

Closes #35

### DIFF
--- a/src/pytom_tm/io.py
+++ b/src/pytom_tm/io.py
@@ -78,9 +78,9 @@ def write_angle_list(data: npt.NDArray[float], file_name: pathlib.Path, order: t
             fstream.write(' '.join([str(x) for x in [data[j, i] for j in order]]) + '\n')
 
 
-def read_mrc_meta_data(file_name: pathlib.Path) -> dict:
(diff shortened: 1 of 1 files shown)


sroet (approved):
LGTM, I would like to also have an added test so that we can read slightly borked meta_data but still give a reasonable error if it is completely useless. But for a quick fix this seems fine

McHaillet:
Agreed, I added an isue for this!

Issue #29, opened 2023-09-01
symmetry option for the template matching search


McHaillet:
Symmetrical particles should requires a reduction in the angular search based on their symmetry

McHaillet:
> LGTM, but we should probably also test symmetric templates at some point (and make sure we get all expected multiple angles with identical scores mentioned in [this comment](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/issues/25#issuecomment-1708111589))

_Originally posted by @sroet in https://github.com/SBC-Utrecht/pytom-template-matching-gpu/pull/40#pullrequestreview-1613149371_

PR #40, opened 2023-09-06
non-symmetrical object solves numerical incosistencies in job splitting unittest

McHaillet:
Unittests for job splitting were not consistent. Making the object non-symmetrical solved the issue on my machine and cluster. The test is even much more concise now.

See issue #25.


sroet (approved):
LGTM, but we should probably also test symmetric templates at some point (and make sure we get all expected multiple angles with identical scores mentioned in [this comment](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/issues/25#issuecomment-1708111589))

McHaillet:
I added it to issue #29 which is about reduced search for symmetrical objects.

PR #42, opened 2023-09-07
Extraction mask feature

### LINKED ISSUE #11: add membrane segmentation input to extract_candidates
Membrane segmentation should be provided as an mrc file with binary annotation of membrane (1) and background (0) that has the same dimensions of the tomogram.

There should be multiple use cases
- search only inside the membrane by applying the mask to the score volume
- search this many voxels outside the memrbane. Dilate the membrane with a number of voxels and apply the dilated mask to the score volume
- search only around the membrane. First dilate membrane with a number of voxels, that subtract the original membrane mask. Finally apply the resulting mask to the score volume.

### LINKED ISSUE #36: Adding tomogram masking option
Hi, 

We are doing a large number of template matching jobs and often use Amira masks to limit particles to be inside the cell. In the previous version there was on option of adding a tomogram mask but it seems to be gone now. 

At the moment we are using imod clip to multiply the scores file to the mask before extraction, but have masks in the input of extraction could be much more convenient. 

Thanks,
Hamid

McHaillet:
This closes #36 and closes #11.

The candidate extraction accepts a mask for a specific region of interest in the tomogram. Aimed at allowing users to extract particles within some cellular context.

Also added tests to ensure masked areas are extracted (or not).

### DIFF
--- a/src/bin/pytom_extract_candidates.py
+++ b/src/bin/pytom_extract_candidates.py
@@ -14,6 +14,11 @@ def main():
     parser.add_argument('-j', '--job-file', type=pathlib.Path, required=True, action=CheckFileExists,
                         help='JSON file that contain all data on the template matching job, written out by '
                              'pytom_match_template.py in the destination path.')
+    parser.add_argument('--tomogram-mask', type=pathlib.Path, required=False, action=CheckFileExists,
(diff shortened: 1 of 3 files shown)


sroet (approved):
LGTM, feel free to merge

PR #43, opened 2023-09-08
Starfile annotation patch

### LINKED ISSUE #20: write out LCCmax scores in pytom_extract_candidates.py
Uncommented this because it was giving errors with reading in other software.

These lines of code:
- https://github.com/SBC-Utrecht/pytom-template-matching-gpu/blob/f775eadb062e369f893541d12bd6cb5ab362f601/src/pytom_tm/extract.py#L99
- https://github.com/SBC-Utrecht/pytom-template-matching-gpu/blob/f775eadb062e369f893541d12bd6cb5ab362f601/src/pytom_tm/extract.py#L121

### LINKED ISSUE #37: Including the LCCmax in the star file
As I mentioned previously in #36 , we often mask our tomograms to limit false positives and this throws off the distribtion of the score (local cross-correlation).  it would be extremely useful for us to have scores of each particle included in the output starfiles so we can manipulate the files later when we go through our picks. 

Thanks in advance,
Hamid

McHaillet:
Draft PR for making the starfiles more descriptive. Closes #37 and closes #20.

NOTE: pytom main needs to be updated to read this format.

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -132,7 +132,7 @@ def extract_particles(
         'ptmAnglePsi',
         'ptmLCCmax',
         'ptmCutOff',
-        'ptmSearchStd'
+        'ptmSearchStd',
         'ptmDetectorPixelSize',
         'ptmMicrographName',
         # 'ptmMagnification',
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -55,9 +55,9 @@ def extract_particles(
         ]
         score_volume[tomogram_mask <= 0] = -1
 
+    sigma = job.job_stats['std']
     if cut_off is None:
         # formula rickgauer (2017) should be: 10**-13 = erfc( theta / ( sigma * sqrt(2) ) ) / 2
-        sigma = job.job_stats['std']
         search_space = (
             # wherever the score volume has not been explicitly set to -1 is the size of the search region
             (score_volume > -1).sum() *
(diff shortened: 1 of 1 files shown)


McHaillet:
@sroet brief explanation of this update.

- Removed two columns (Magnification and GroupNumber) which I filled with default values as I thought it was neccessary for reading in Warp. That is not the case, hence why I removed them.
- Added three columns with extraction statistics, the score of the particle, the extraction threshold, and the standard deviation of the search.
- Replaced _rln with _ptm for the STAR file header. That way its recognizable that the results are calculated in this package. For reading in Warp (and perhaps other software), this might mean running `sed -i 's/_ptm/_rln' file.star`. I can add this in the wiki instructions. What is your opinion on this change?

sroet (approved):
LGTM feel free to merge

> * Removed two columns (Magnification and GroupNumber) which I filled with default values as I thought it was neccessary for reading in Warp. That is not the case, hence why I removed them.
> 
> * Added three columns with extraction statistics, the score of the particle, the extraction threshold, and the standard deviation of the search.
> 

thanks for the added comments 😄 


> * Replaced _rln with _ptm for the STAR file header. That way its recognizable that the results are calculated in this package. For reading in Warp (and perhaps other software), this might mean running `sed -i 's/_ptm/_rln' file.star`. I can add this in the wiki instructions. What is your opinion on this change?

No strong feelings either way, so let's stick with what you feel is best. Please do update the wiki if you decide to change it to `_ptm`

McHaillet:
> Please do update the wiki if you decide to change it to `_ptm`

Will do!

Issue #46, opened 2023-10-09
invalid resolution specified


DrJesseHansen:
hello, 

thanks for the great tools! 

I am trying to create a template.  I have used a PDB (and done molmap) and also tried downloading an EM density and both times I get the same error.  

First, when I check the header of my mrc input I get:
Pixel spacing (Angstroms)..............   1.693      1.693      1.693 

Now my command:
pytom_create_template.py -i emd_1.mrc -o template.mrc --input-voxel-size-angstrom 1.693 --output-voxel-size-angstrom 13.54

WARNING:root:Provided voxel size does not match voxel size annotated in input map.
WARNING:root:Invalid resolution specified, changing to 27.08A
INFO:root:Convoluting volume with filter and then downsampling.

No idea what's happening here.  Thanks in advance

Jesse

McHaillet:
Hi Jesse,

I was on holidays the last weeks, so getting to your issue now.

If you run the following on your template with the conda environement active, 

``` python -c "from pytom_tm.io import read_mrc_meta_data; print(read_mrc_meta_data('emd_1.mrc')['voxel_size'])``` ,

you will probably see some small decimal point values that are not significant. So, you can safely skip the warning.

Anyway I will push a fix to make the warning a bit less stringent because it was giving errors before in another part of the code!

Best,
Marten

**Comment by McHaillet** [comment]
Both warnings have been solved with PR #47.

You can update your installation with the following. First activate your environment:

```
conda activate pytom_tm
```

Then go to the folder where you cloned the repo:

```
cd pytom-template-matching-gpu
```

Pull latest changes and reinstall:

```
git pull
pip install .[plotting]
```

DrJesseHansen:
hi, just wanted to update that I had our IT dept install the latest version (I guess  pytom-template-matching-gpu/0.2-

pytom_create_template.py -i input_map.mrc -o template.mrc --input-voxel-size-angstrom 2.762 --output-voxel-size-angstrom 11.048 --center -c -z 5 -a 0.07 -v 300 --Cs 2.7 --cut-after-first-zero
WARNING:root:Resolution is either not specified (in which case you can ignore this warning) or incorrectly specified, changing to 22.096A
INFO:root:Convoluting volume with filter and then downsampling.

McHaillet:
Hi, sorry I do not understand what the issue is now. You are getting an output volume, right?

[...]

PR #47, opened 2023-10-18
Update pytom_create_template.py

McHaillet:
Solves issue #46.

I made the voxel size check a bit less stringent (similar to how this was an issue before in the template matching code).

Secondly, I made the warning for resolution more clear.

### DIFF
--- a/src/bin/pytom_create_template.py
+++ b/src/bin/pytom_create_template.py
@@ -68,7 +68,7 @@ def main():
     input_data = read_mrc(args.input_map)
     input_meta_data = read_mrc_meta_data(args.input_map)
     if args.input_voxel_size_angstrom is not None:
-        if args.input_voxel_size_angstrom != input_meta_data['voxel_size']:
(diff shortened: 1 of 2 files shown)


src/bin/pytom_create_template.py
```diff
@@ -68,7 +68,7 @@ def main():
     input_data = read_mrc(args.input_map)
     input_meta_data = read_mrc_meta_data(args.input_map)
     if args.input_voxel_size_angstrom is not None:
-        if args.input_voxel_size_angstrom != input_meta_data['voxel_size']:
+        if args.input_voxel_size_angstrom != round(input_meta_data['voxel_size'], 3):
```
sroet, line 71: for sanity I would round both sides:
```suggestion
        if round(args.input_voxel_size_angstrom, 3) != round(input_meta_data['voxel_size'], 3):
```

sroet (changes requested):
1 small change, LGTM otherwise

sroet (approved):
LGTM, feel free to merge

PR #51, opened 2023-10-24
Update warning for "no filter resolution"

### LINKED ISSUE #46: invalid resolution specified

sroet:
Closes #46 by updating the warning to be more verbose about which resolution is not given

### DIFF
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -37,8 +37,8 @@ def generate_template_from_map(
         )
 
     if filter_to_resolution is None or filter_to_resolution < (2 * output_spacing):
-        logging.warning(f'Resolution is either not specified (in which case you can ignore this warning) '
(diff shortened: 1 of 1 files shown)


McHaillet (approved):
Great! Thanks for updating :)

PR #55, opened 2023-11-01
reduce angular search space by applying symmetry

### LINKED ISSUE #29: symmetry option for the template matching search

McHaillet:
Closes #29 

I added a small option to reduce angular search space for n-fold symmetrical particles.
- pytom_match_template.py has a symmetry option
- the symmetry is applied in TMJob by indexing into the angular search lists

Can this still be part of the 0.3.0 version or should I increase th version number?

### DIFF
--- a/src/bin/pytom_match_template.py
+++ b/src/bin/pytom_match_template.py
@@ -38,9 +38,9 @@ def main():
     parser.add_argument('--angular-search', type=str, required=True,
                         help='Options are: [7.00, 35.76, 19.95, 90.00, 18.00, '
                              '12.85, 38.53, 11.00, 17.86, 25.25, 50.00, 3.00]')
-    parser.add_argument('--symmetry', type=int, required=False, action=LargerThanZero, default=1,
-                        help='Reduce the angular search for symmetrical particles. The length of the rotation '
-                             'search will be shortened through division by this value.')
+    parser.add_argument('--rotational-symmetry', type=int, required=False, action=LargerThanZero, default=1,
+                        help='Integer value indicating the rotational symmetry of the template. The length of the '
+                             'rotation search will be shortened through division by this value.')
     parser.add_argument('-s', '--volume-split', nargs=3, type=int, required=False, default=[1, 1, 1],
                         help='Split the volume into smaller parts for the search, can be relevant if the volume does '
                              'not fit into GPU memory. Format is x y z, e.g. --volume-split 1 2 1')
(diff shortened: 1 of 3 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -303,7 +305,7 @@ def merge_sub_jobs(self, stats: Optional[list[dict, ...]] = None) -> tuple[npt.N
             return result
 
         if stats is not None:
-            search_space = reduce(lambda x, y: x * y, self.search_size) * self.n_rotations
+            search_space = reduce(lambda x, y: x * y, self.search_size) * - (-self.n_rotations // self.symmetry)
```
sroet, line 308: what is up with the double negative here?
If this is a trick to round up, please use `from math import ceil` and `ceil(self.n_rotations / self.symmetry)`

src/pytom_tm/tmjob.py
```diff
@@ -430,8 +432,12 @@ def start_job(
                 )
 
         # load rotation search
-        angle_ids = list(range(self.start_slice, self.n_rotations, self.steps_slice))
-        angle_list = load_angle_list(self.rotation_file)[slice(self.start_slice, self.n_rotations, self.steps_slice)]
+        angle_ids = list(range(self.start_slice, - (-self.n_rotations // self.symmetry), self.steps_slice))
```
sroet, line 435: same as above with the double negative

src/pytom_tm/tmjob.py
```diff
@@ -430,8 +432,12 @@ def start_job(
...
+        angle_list = load_angle_list(self.rotation_file)[slice(
+            self.start_slice,
+            - (-self.n_rotations // self.symmetry),
```
sroet, line 438: same as above with the double negative

sroet (changes requested):
Please use slightly less obscure ways to round up 😅 

> Can this still be part of the 0.3.0 version or should I increase th version number?

I don't see breaking changes, so feel free to keep the version number as is. Also, normally the version is only "locked-in" after a [proper release has been done](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/releases). So no need to update it further (even for breaking changes) untill we do another proper release (by tagging it and drafting it in [here](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/releases/new))

src/bin/pytom_match_template.py
```diff
@@ -38,6 +38,9 @@ def main():
     parser.add_argument('--angular-search', type=str, required=True,
                         help='Options are: [7.00, 35.76, 19.95, 90.00, 18.00, '
                              '12.85, 38.53, 11.00, 17.86, 25.25, 50.00, 3.00]')
+    parser.add_argument('--symmetry', type=int, required=False, action=LargerThanZero, default=1,
+                        help='Reduce the angular search for symmetrical particles. The length of the rotation '
+                             'search will be shortened through division by this value.')
```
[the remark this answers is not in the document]
McHaillet: Probably yes, but I am not so familiar with this. I now updated everything to rotational_symmetry to be more specific about the symmetry type of the template that it is applicable for.

sroet (approved):
LGTM, feel free to merge

PR #56, opened 2023-11-01
Allow custom angular sampling

McHaillet:
This is a small change (and the last for now) to allow users to provide their own set of rotations to search the tomogram with.


sroet (changes requested):
LGTM, but please add a test so that this option stays available in future development

McHaillet:
good point, added it in 👍

sroet (approved):
LGTM feel free to merge

PR #57, opened 2023-11-02
version 0.3.1: job annotation with module version number, symmetry patch, and whitening filter calculation check

McHaillet:
Realised that I forgot to apply the symmetry in extraction threshold estimation where it should be used to calculate the full search space.

MUCH more importantly: realised that the angle list was not yet sort and needs to be sorted for effective symmetry reduction!
They are now sorted by the first euler angle upon loading, which should be a general fix that also works for user input angular search.


McHaillet:
For some reason this broke something in the unittests. Need to investigate.

McHaillet:
These tests are currently failing: 

```
======================================================================
FAIL: test_search (test_template_matching.TestTM.test_search)
----------------------------------------------------------------------
  File "/home/mchaillet/projects/pytom-template-matching-gpu/tests/test_template_matching.py", line 43, in test_search
    self.assertAlmostEqual(stats['std'], 0.005175, places=6,
AssertionError: 0.005174099103974734 != 0.005175 within 6 places (9.008960252659959e-07 difference) : Standard deviation of the search should be almost equal

======================================================================
FAIL: test_parallel_manager (test_tmjob.TestTMJob.test_parallel_manager)
----------------------------------------------------------------------
  File "/home/mchaillet/projects/pytom-template-matching-gpu/tests/test_tmjob.py", line 220, in test_parallel_manager
    self.assertTrue(score.max() > 0.934, msg='lcc max value lower than expected')
AssertionError: False is not true : lcc max value lower than expected

======================================================================
FAIL: test_tm_job_split_angles (test_tmjob.TestTMJob.test_tm_job_split_angles)
----------------------------------------------------------------------
  File "/home/mchaillet/projects/pytom-template-matching-gpu/tests/test_tmjob.py", line 207, in test_tm_job_split_angles
    self.assertTrue(score.max() > 0.934, msg='lcc max value lower than expected')
AssertionError: False is not true : lcc max value lower than expected

======================================================================
FAIL: test_tm_job_split_volume (test_tmjob.TestTMJob.test_tm_job_split_volume)
----------------------------------------------------------------------
  File "/home/mchaillet/projects/pytom-template-matching-gpu/tests/test_tmjob.py", line 175, in test_tm_job_split_volume
    self.assertTrue(score.max() > 0.934, msg='lcc max value lower than expected')
AssertionError: False is not true : lcc max value lower than expected

----------------------------------------------------------------------
Ran 18 tests in 9.616s

FAILED (failures=4)
```

- For the first one I think its safe to reduce to places=6 as it is a very minimal difference.
- For the others, the correlation went down from 0.934 to 0.932 (rounded). Not entirely sure what the issue is. Could it be because the rng becomes different now that the [angle loading code slightly changed](https://github.com/McHaillet/pytom-template-matching-gpu/blob/402636fe5208f7b8a83b94f417c742ba5cb760f2/tests/test_tmjob.py#L55)?

sroet (approved):
LGTM in general, will approve

The API did not change, so a bugfix version change seems appropriate.

I am not too sure if I understand what is incompatible between the two version. Is it just that the result will be different, or can they not be loaded/analyzed at all?

McHaillet:
For my part ready to go. I made the following further updates:

Okay, the version number is now a variable stored in the TMJob structure. When a job is initialized the current module version number is assigned, through `pytom_tm.__version__` (which I had to add). When jobs without version numbers are loaded from a json, it must be an older job and therefore they are assigned version '0.3.0', with data.get('version', '0.3.0').

I added a `sort_angles` option to the angle list loading and in the code this option is True when the job version number is larger than '0.3.0'. I used the version module form packaging to make the check.

For the whitening filter, I added a check to see if the file exists upon job init. This is to prevent recalculation of the filter upon loading a job for particle extraction as it is quite an expensive operation.

setup.py
```diff
@@ -1,5 +1,8 @@
 import setuptools
 
+# get the version number
+exec(open('src/pytom_tm/_version.py').read())
+
```
sroet, line 5: I really dislike using `exec`, also [using `__version__` inside packages is a rejected PEP](https://peps.python.org/pep-0396/#pep-rejection).
[This answer](https://stackoverflow.com/a/73119016/21764481) would be more suitable, so hardcode the version in `setup.py` (this allows us to use configure files later on) and then using 
```python
from importlib.metadata import version
version = version('pytom_tm') #guessing the name here
``` 
as a version file or inside `src/pytom_tm/__init__.py`

setup.py
```diff
@@ -8,7 +11,7 @@
     name='pytom-template-matching-gpu',
     packages=['pytom_tm', 'pytom_tm.angle_lists'],
     package_dir={'': 'src'},
-    version='0.3.0',  # for versioning definition see https://semver.org/
+    version=__version__,  # for versioning definition see https://semver.org/
```
sroet, line 14: see above

src/pytom_tm/__init__.py
```diff
@@ -1,3 +1,5 @@
+from ._version import __version__
+
```
sroet, line 2: see above

src/pytom_tm/angles.py
```diff
@@ -22,13 +22,15 @@
...
+def load_angle_list(file_name: pathlib.Path, sort_angles: bool = True) -> list[tuple[float, float, float]]:
     with open(str(file_name)) as fstream:
         lines = fstream.readlines()
     angle_list = [tuple(map(float, x.strip().split(' '))) for x in lines]
     if not all([len(a) == 3 for a in angle_list]):
         raise ValueError('Invalid angle file provided, each line should have 3 ZXZ Euler angles!')
     else:
+        if sort_angles:
+            angle_list.sort()  # angle list needs to be sorted otherwise symmetry reduction cannot be used!
         return angle_list
```
sroet, line 34: `else` not needed after raising an error
```suggestion
    if sort_angles:
        angle_list.sort()  # angle list needs to be sorted otherwise symmetry reduction cannot be used!
    return angle_list
```

src/pytom_tm/tmjob.py
```diff
@@ -1,4 +1,6 @@
 from __future__ import annotations
+from pytom_tm import __version__
```
sroet, line 2: see above

sroet (changes requested):
You are using a very old/insecure solution for the versioning, please update to use `importlib.metadata.version` instead

McHaillet:
I switched everything to importlib!

`__init__.py` now also defines `__version__` with this method.

sroet (approved):
Thanks for the changes, LGTM, feel free to merge!

PR #60, opened 2023-11-08
Patch for extraction cut off

McHaillet:
I encountered an error where `pytom_extract_candidates.py` was crashing because it was trying to extract a particle at the edge of the score volume. This should never happen as in line 46-51 the edge region is masked out.

- Regions of the score map that should not be extracted are now set to 0 instead of -1. This was leading to errors as the peak mask that was multiplied with the score map had 0 values, which lead to -1 values being reset to 0.
- This now also means the extraction cut_off cannot be smaller than 0 and the check for each particle (in extract.py line 92) is now changed to `<=` instead of `<` as this guarantees the program stops once a 0 value has been encountered as the maximum.

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -43,12 +43,12 @@ def extract_particles(
     )
 
     # mask edges of score volume
-    score_volume[0: particle_radius_px, :, :] = -1
-    score_volume[:, 0: particle_radius_px, :] = -1
-    score_volume[:, :, 0: particle_radius_px] = -1
-    score_volume[-particle_radius_px:, :, :] = -1
-    score_volume[:, -particle_radius_px:, :] = -1
-    score_volume[:, :, -particle_radius_px:] = -1
+    score_volume[0: particle_radius_px, :, :] = 0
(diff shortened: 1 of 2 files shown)


src/pytom_tm/extract.py
```diff
@@ -69,6 +69,8 @@ def extract_particles(
         )
         cut_off = erfcinv((2 * n_false_positives) / search_space) * np.sqrt(2) * sigma
         logging.info(f'cut off for particle extraction: {cut_off}')
+    elif cut_off < 0:
+        cut_off = 0
```
sroet, line 73: this should give a warning (as we might be silently overriding user settings here) or input sanitation has to happen in the arparse

sroet (changes requested):
1 change and please update the version number for this bugfix

src/pytom_tm/extract.py
McHaillet: Added a logging.warning for this one

McHaillet:
version is also updated to 0.3.2

sroet (approved):
LGTM, feel free to merge

Issue #61, opened 2023-11-08
ValueError: Either low-pass or high-pass needs to be set for band-pass


Phaips:
Hey,

I get an error testing the new version with the full 3D CTF and wedge model.
My inputs are:

pytom_match_template.py -t /../template_tomo_612.mrc --mask mask.mrc -v tomo/612.rec -d /612 -a temp/612.tlt -s 2 2 1 --angular-search 7.00 -g 0 1 2 3 --voxel-size-angstrom 7.84 --non-spherical-mask --per-tilt-weighting --amplitude-contrast 0.07 --dose-accumulation temp/612_exposure.txt --defocus-file temp/612_defocus.txt --voltage 300 --spherical-abberation 2.7

and the error comes from the weights.py

```
/../miniconda3/envs/pytom_tm_v0.3.1/lib/python3.9/site-packages/pytom_tm/weights.py", line 131, in create_gaussian_band_pass
    raise ValueError('Either low-pass or high-pass needs to be set for band-pass')
ValueError: Either low-pass or high-pass needs to be set for band-pass
```

Since we never lp or hp filtered our tomos and never had this issue in version 0.2 I am not sure where the error is coming from.

McHaillet:
Thanks for the report! Will try to make a fix for that ASAP

McHaillet:
And btw, setting the low-pass to the nyquist resolution should already allow you to run it right now without removing signal.

Phaips:
> And btw, setting the low-pass to the nyquist resolution should already allow you to run it right now without removing signal.

Thanks for the quick reply! This is already a great solution and it works!

McHaillet:
Issue has been solved with PR #62 !

PR #62, opened 2023-11-09
Patch band pass unassigned

McHaillet:
This PR will be part of the bugfixes for version 0.3.2. It solves issue #61.

I updated the tomogram (`tomogram_filter`) and template weighting generation (`template_wedge`) in `TMJob.start_job()` to solve this issue. After the addition of the spectrum whitening there are now quite some weighting options: a band-pass, a whitening-filter, and a wedge filter. They are optional but should also work together.

To make it more logical both weighting are now initialized as 1, and are multiplied with all the optional filters in when the respective flags are set. If they remain 1 (no filter flags are set) they will not modify the tomogram/template.

I updated the tests to run the job with combinations of these weighting options.

Also removed some debug volume writing as it interfered with clean up of the test data now that I made the filter creation more logical.

I am a bit uncertain about the fact that in `test_weights.py` I have some variables that store general data (CTF, dose accumulation) and load this from this file into `test_tmjob.py`, is that okay to do?

### DIFF
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -393,61 +393,59 @@ class TMJob:
             read_mrc(self.mask)
         )
 
-        # create weighting, include missing wedge and band-pass filters
-        template_wedge = None
-        if self.tilt_angles is not None:
-            # convolute tomo with whitening filter and bandpass
-            tomo_filter = (create_gaussian_band_pass(
-                search_volume.shape,
-                self.voxel_size, 
+        # init tomogram and template weighting
+        tomo_filter, template_wedge = 1, 1
+        # first generate bandpass filters
+        if not (self.low_pass is None and self.high_pass is None):
+            tomo_filter *= create_gaussian_band_pass(
+                    search_volume.shape,
+                    self.voxel_size,
+                    self.low_pass,
+                    self.high_pass
+            ).astype(np.float32)
+            template_wedge *= create_gaussian_band_pass(
+                self.template_shape,
+                self.voxel_size,
                 self.low_pass,
                 self.high_pass
-            ) * (profile_to_weighting(
+            ).astype(np.float32)
+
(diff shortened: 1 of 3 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -393,61 +393,59 @@ def start_job(
...
+        if self.tilt_angles is not None:
+            # for the tomogram a binary wedge is generated to explicitly set the missing wedge region to 0
+            tomo_filter *= create_wedge(
+                search_volume.shape,
+                self.tilt_angles,
+                self.voxel_size,
+                cut_off_radius=1.,
+                angles_in_degrees=True,
+                tilt_weighting=False
+            )
```
sroet, line 434: is there a reason this is not `astype`'d?
```suggestion
            tomo_filter *= create_wedge(
                search_volume.shape,
                self.tilt_angles,
                self.voxel_size,
                cut_off_radius=1.,
                angles_in_degrees=True,
                tilt_weighting=False
            ).astype(np.float32)
```

tests/test_tmjob.py
```diff
@@ -10,6 +10,7 @@
 from pytom_tm.tmjob import TMJob, TMJobError
 from pytom_tm.io import read_mrc, write_mrc, UnequalSpacingError
 from pytom_tm.extract import extract_particles
+from test_weights import CTF_PARAMS, ACCUMULATED_DOSE, TILT_ANGLES
```
sroet, line 13: this is fine, often for bigger programming suites we have a dedicated `common.py` file inside the test directory to define these parameters and import the from there.
Another option is a default `class` in this common file that sets all these options on initialization and all the other tests inherit from that one

sroet (changes requested):
one missed conversion LGTM otherwise

sroet (approved):
LGTM, feel free to merge

PR #63, opened 2023-11-09
improve radial average speed with nimdage.mean()

McHaillet:
One more update for version 0.3.2. I discovered online that the radial average calculation could be much faster using ndimage.mean().

Results completely overlap but the speed increase is massive:

whitening_filter

### DIFF
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -575,7 +575,8 @@ def power_spectrum_profile(image: npt.NDArray[float]) -> npt.NDArray[float]:
         axes=(0, 1) if len(image.shape) == 3 else 0
     )
 
-    return ndimage.mean(power, labels=q_grid, index=q)
+    power_profile = ndimage.mean(power, labels=q_grid, index=q)
+    return power_profile
 
 
 def profile_to_weighting(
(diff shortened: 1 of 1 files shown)


src/pytom_tm/weights.py
```diff
@@ -565,18 +565,17 @@ def power_spectrum_profile(image: npt.NDArray[float]) -> npt.NDArray[float]:
...
+    q_grid = np.floor(
+        # convert to radial indices in the fourier power spectrum, 0.5 is added to obtain the correct ring
+        radial_reduced_grid(image.shape) * (max(image.shape) // 2) + 0.5
+    ).astype(int)
+    power = np.fft.fftshift(
+        np.abs(np.fft.rfftn(image)) ** 2,
+        axes=(0, 1) if len(image.shape) == 3 else 0
+    )
+
+    return ndimage.mean(power, labels=q_grid, index=q)
```
sroet, line 578: I would prefer an assignment before the return
```suggestion
    power_profile = ndimage.mean(power, labels=q_grid, index=q)
    return power_profile
```

sroet (approved):
1 small comment, feel free to ignore. LGTM otherwise.
Also, please document the "massive speed increase" are we talking about 2x, 10x 10000x?

McHaillet:
> 1 small comment, feel free to ignore. LGTM otherwise. Also, please document the "massive speed increase" are we talking about 2x, 10x 10000x?

This is for a large tomogram, dimensions 900x900x300.
Old version took ~425s, new version ~16s. So the speed up is roughly 25x

PR #65, opened 2023-11-13
update to minimal cutoff value in estimate_roc

McHaillet:
The change of minimal cutoff value to 0 in version 0.3.2, was not yet updated in estimate ROC. 

(sorry for commit directly to main)


sroet (approved):
seems like an easy fix, LGTM feel free to merge

PR #66, opened 2023-11-16
Patch unittest inconsistency

### LINKED ISSUE #58: error message of the unittests
Hi there,

I am trying to install and test pytom-template matching on my own workstation, with unbuntu 22.04, nvidia 4090 and cuda 12.2.

No error message during the installation. However, I got the following error message (see the attached snapshot) for the unittests. Anyone know how to fix the issue?

Thanks in advance!
Yunjie
Screenshot from 2023-11-03 22-00-21

McHaillet:
Closes #58. 

It works! :relieved:

- Both the test_parallel_manager and test_tm_job_split_volume now only compare the score and angle volume with a distance of template_size // 2 from the edge as we expect numerical inaccuracies in those regions.
- test_tm_job_split_volume now splits along each dimensions (2,3,2) creating 12 subvolumes 
- test_parallel_manager still only splits (1,3,1) as this test should just check the parallel managing part

### DIFF
--- a/tests/test_tmjob.py
+++ b/tests/test_tmjob.py
@@ -220,18 +220,15 @@ class TestTMJob(unittest.TestCase):
         # is not well defined in the boundary area, only a small part of the template is correlated here (and we are
         # not really interested in it). Probably the inaccuracy in this area becomes more apparent when splitting
         # into subvolumes due to a smaller number of sampling points in Fourier space.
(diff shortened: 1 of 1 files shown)


McHaillet:
Added a test that checks whether split volume raises the correct error with too small pieces

sroet (changes requested):
Small readability improvement, LGTM othewise

McHaillet:
Added your suggestions and checked that the tests still succeed!

sroet (approved):
LGTM, feel free to merge

PR #68, opened 2023-12-12
Bump version  to 0.3.3

McHaillet:
PR to make sure the lastest version can be tagged under a new version number. Since the last version we patched the following:

- infamous unittest bug
- radial average calculation unified to single function
- minimal cutoff in estimate_roc now set correctly


sroet (approved):
LGTM, feel free to merge

PR #69, opened 2023-12-12
Improved installation instructions

thorstenwagner:
Thanks for this package :-) 

I would like to suggest some changes:

1. Install cupy via conda-forge, otherwise it will try to install it from source, which often fails.
2. When creating the cuda env, base it on cuda 11.8. This way your users are not forced to have the latest cuda driver installed.
3. Improved compatibility with zsh: `pip install '.[plotting]'`. Without the quotes it will crash with zsh.

Best,
Thorsten

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,14 @@ Create a new conda environment:
 conda create -n pytom_tm -c conda-forge python=3 cupy cuda-version=11.8
 ```
 
+(2) Create an environment without cupy and let pip build against a system installed CUDA-toolkit:
+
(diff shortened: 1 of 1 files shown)


McHaillet (changes requested):
Hi Thorsten,

Thanks a lot for the PR for improved install instructions!

A few suggested changes:
-  I think its a good idea to change the default to the prebuild cupy with cuda-toolkit from conda-forge, but also wanted to offer the cupy build if people are comfortable with that.
- A few things I now realised that might be good to add in ;)

Best!

Issue #74, opened 2024-01-05
ERROR: test_parallel_manager (test_tmjob.TestTMJob.test_parallel_manager)


chris-hypercag:
Hello developers,

I was able to install the software successfully per the README instructions. However when I ran the unittest I encountered an error with `test_parallel_manager`. The error is similar to issue [#58](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/issues/58) and [#66](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/pull/66), except that the traceback for the error is different. 

```
======================================================================
ERROR: test_parallel_manager (test_tmjob.TestTMJob.test_parallel_manager)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../pytom-template-matching-gpu/tests/test_tmjob.py", line 262, in test_parallel_manager
    score, angle = run_job_parallel(self.job, volume_splits=(1, 3, 1), gpu_ids=[0])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../anaconda3/envs/pytom_tm/lib/python3.12/site-packages/pytom_tm/parallel.py", line 96, in run_job_parallel
    raise RuntimeError('One of the processes stopped unexpectedly.')
RuntimeError: One of the processes stopped unexpectedly.

----------------------------------------------------------------------
```

I am running the software on a large heterogeneous cluster, and ran into the error when executing the test with a Nvidia Tesla V100-SXM2, and Tesla P40. The error did not occur when running on an Nvidia A30.  All three of these GPU's are currently supporting CUDA version 12.3 with driver version 545.23.08. 

In all three cases running `python -c "import cupy as cp; a = cp.zeros((100,100)); print(a)"` produced the correct output, so I told the user they can probably use the software, just to be wary of the results. I've attached the entire test output; it contains more information than the excerpt above. Please, let me know if you need anymore info.

Cheers,
Chris

[github_issue_report.txt](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/files/13845310/github_issue_report.txt)

McHaillet:
Hi Chris,

Thanks for the detailed report. Your issue seems different from #58 and #66 as those had a numerical incosistency in the tests. In your case it seems rather that the test broke because at one point the GPU was unavailable (line 33 in your attached file):

[...]

PR #75, opened 2024-01-09
Terminate other processes if one breaks

McHaillet:
Noticed a multiprocessing problem (unrelated to #74). When one of the processes fails, the other will keep running. This is undesired as the results will become unusable. Now I explicitly terminate all processes when one fails before raise a runtimeerror in the main process.

(I previsouly thought that 'raise RuntimeError' would be propagated to child processes, but that is apparently not the case)

### DIFF
--- a/src/pytom_tm/parallel.py
+++ b/src/pytom_tm/parallel.py
@@ -3,10 +3,11 @@ import multiprocessing as mp
 import logging
 import queue
 import time
-import sys, os
+import contextlib
 from multiprocessing.managers import BaseProxy
 from functools import reduce
 from pytom_tm.tmjob import TMJob
+from pytom_tm.utils import mute_stdout_stderr
 
 try:  # new processes need to be spawned in order to set cupy to use the correct GPU
     mp.set_start_method('spawn')
@@ -18,16 +19,17 @@ def gpu_runner(
         gpu_id: int, task_queue: BaseProxy, result_queue: BaseProxy, log_level: int, mute: bool
 ) -> None:
     if mute:
-        devnull = open(os.devnull, 'w')
-        sys.stdout = devnull
-        sys.stderr = devnull
-    logging.basicConfig(level=log_level)
-    while True:
-        try:
-            job = task_queue.get_nowait()
-            result_queue.put_nowait(job.start_job(gpu_id, return_volumes=False))
-        except queue.Empty:
-            break
+        mute_context = mute_stdout_stderr()  # Make sure it is imported at this time
+    else:
+        mute_context = contextlib.nullcontext()
+    with mute_context:
+        logging.basicConfig(level=log_level)
+        while True:
+            try:
+                job = task_queue.get_nowait()
+                result_queue.put_nowait(job.start_job(gpu_id, return_volumes=False))
+            except queue.Empty:
+                break
 
 
 def run_job_parallel(
--- a/src/pytom_tm/parallel.py
+++ b/src/pytom_tm/parallel.py
@@ -39,7 +39,7 @@ def run_job_parallel(
     @param main_job: a TMJob object from pytom_tm that contains all data for a search
     @param volume_splits: tuple of len 3 with splits in x, y, and z
     @param gpu_ids: list of gpu indices available for the job
-    @param mute: boolean to mute spawned process terminal output, only used for unittesting
+    @param unittest_mute: boolean to mute spawned process terminal output, only used for unittesting
     @return: the volumes with the LCCmax and angle ids
     """
 
--- a/src/pytom_tm/parallel.py
+++ b/src/pytom_tm/parallel.py
@@ -3,6 +3,7 @@ import multiprocessing as mp
 import logging
 import queue
 import time
+import sys, os
 from multiprocessing.managers import BaseProxy
 from functools import reduce
 from pytom_tm.tmjob import TMJob
@@ -13,7 +14,13 @@ except RuntimeError:
     pass
 
 
-def gpu_runner(gpu_id: int, task_queue: BaseProxy, result_queue: BaseProxy, log_level: int) -> None:
+def gpu_runner(
+        gpu_id: int, task_queue: BaseProxy, result_queue: BaseProxy, log_level: int, mute: bool
+) -> None:
+    if mute:
+        devnull = open(os.devnull, 'w')
+        sys.stdout = devnull
+        sys.stderr = devnull
     logging.basicConfig(level=log_level)
     while True:
         try:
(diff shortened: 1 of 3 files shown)


sroet (changes requested):
LGTM, as discussed offline: please add a test that catches this `RuntimeError` and then checks if there are no  processes still in `multiprocess.active_children()`

McHaillet:
Not done yet, will let you know when ready ;)

McHaillet:
Okay, good to go for my part!

I added in a stdout/stderr muting option for spawning processes during unit tests to prevent the terminal showing an error message during the tests. Let me know if you agree with how I did that.

src/pytom_tm/parallel.py
```diff
@@ -13,7 +14,13 @@
...
 
 
-def gpu_runner(gpu_id: int, task_queue: BaseProxy, result_queue: BaseProxy, log_level: int) -> None:
+def gpu_runner(
+        gpu_id: int, task_queue: BaseProxy, result_queue: BaseProxy, log_level: int, mute: bool
+) -> None:
+    if mute:
+        devnull = open(os.devnull, 'w')
+        sys.stdout = devnull
+        sys.stderr = devnull
```
sroet, line 23: this breaks any further writes to `sys.stdout` and `sys.stderr` by any further code.
it is probably better to make a mute `contextmanager` (following [this so answer](https://stackoverflow.com/a/52442331/21764481)) :
```python
from contextlib import contextmanager,redirect_stderr,redirect_stdout
from os import devnull

@contextmanager
def mute_stdout_stderr():
    """A context manager that redirects stdout and stderr to devnull"""
    with open(devnull, 'w') as fnull:
        with redirect_stderr(fnull) as err, redirect_stdout(fnull) as out:
            yield (err, out)
```
And then in this code:
```python
if mute:
    mute_context = mute_stdout_stderr() #Make sure it is imported at this time
else:
    mute_context =  contextlib.nullcontext()
with mute_context:
    logging.basicConfig(level=log_level)
    while True:
        ...
```

tests/test_tmjob.py
```diff
@@ -258,6 +261,14 @@ def test_tm_job_split_angles(self):
         self.assertTrue(np.abs(angle - read_mrc(TEST_ANGLES)).sum() == 0,
                         msg='split rotation search should be identical')
 
+    def test_parallel_breaking(self):
+        try:
+            _ = run_job_parallel(self.job, volume_splits=(1, 2, 1), gpu_ids=[0, -1], mute=True)
+        except RuntimeError:
+            self.assertEqual(len(multiprocessing.active_children()), 0,
+                             msg='a process was still lingering after a parallel job with partially invalid resources '
+                                 'was started')
```
sroet, line 270: please add an `else` in case the error did not get triggered somehow
```suggestion
        except RuntimeError:
            self.assertEqual(len(multiprocessing.active_children()), 0,
                             msg='a process was still lingering after a parallel job with partially invalid resources '
                                 'was started')
        else:
            self.fail('This should have given a RuntimeError')   
```

sroet (changes requested):
Couple implementation issues, looks good otherwise

src/pytom_tm/parallel.py
McHaillet: that is much nicer, but I am still getting this big error because of the invalid device I am passing

```
Process Process-3:
  File "/home/marten/miniconda3/envs/pytom_tm/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/marten/miniconda3/envs/pytom_tm/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/marten/coding/pytom-template-matching-gpu/src/pytom_tm/parallel.py", line 30, in gpu_runner
    result_queue.put_nowait(job.start_job(gpu_id, return_volumes=False))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marten/coding/pytom-template-matching-gpu/src/pytom_tm/tmjob.py", line 467, in start_job
    tm = TemplateMatchingGPU(
         ^^^^^^^^^^^^^^^^^^^^
  File "/home/marten/coding/pytom-template-matching-gpu/src/pytom_tm/matching.py", line 74, in __init__
    cp.cuda.Device(device_id).use()
  File "cupy/cuda/device.pyx", line 192, in cupy.cuda.device.Device.use
  File "cupy/cuda/device.pyx", line 198, in cupy.cuda.device.Device.use
  File "cupy_backends/cuda/api/runtime.pyx", line 375, in cupy_backends.cuda.api.runtime.setDevice
  File "cupy_backends/cuda/api/runtime.pyx", line 144, in cupy_backends.cuda.api.runtime.check_status
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInvalidDevice: invalid device ordinal
```

sroet: Yeah, reading up on it it seems like that function does explicitly not mute raised errors. 
You can write your own context as a class with `__enter__` and `__exit__` functions and use that. Currently on mobile, but let me know if you want me to write a complete example class (will probably then do it tomorrow)

McHaillet: I pushed a commit with a context manager like this (took some inspiration from an answer in the SO thread). Seems to work now. I left out all os.dup2 that this answer did (https://stackoverflow.com/a/50438156), any idea what that does?

McHaillet: I think as I have it now it should be fine for the tests

src/pytom_tm/parallel.py
```diff
@@ -3,33 +3,43 @@
...
 
 def run_job_parallel(
-        main_job: TMJob, volume_splits: tuple[int, int, int], gpu_ids: list[int, ...]
+        main_job: TMJob, volume_splits: tuple[int, int, int], gpu_ids: list[int, ...], unittest_mute: bool = False,
 ) -> tuple[npt.NDArray[float], npt.NDArray[float]]:
     """
     @param main_job: a TMJob object from pytom_tm that contains all data for a search
     @param volume_splits: tuple of len 3 with splits in x, y, and z
     @param gpu_ids: list of gpu indices available for the job
+    @param mute: boolean to mute spawned process terminal output, only used for unittesting
```
sroet, line 42:
```suggestion
    @param unittest_mute: boolean to mute spawned process terminal output, only used for unittesting
```

sroet (changes requested):
One last nitpick for the documentation, LGTM otherwise

McHaillet:
Thanks for spotting!

PR #78, opened 2024-01-19
parallel manager in a separate test file

McHaillet:
split parallel manager tests to a separate file to potentially prevent exclusive GPU process errors

PR #79, opened 2024-01-23
unittest discover requirest testfiles to start with 'test'

sroet:
#78 fixed the tests suite but dropped the test from discovery (it searches for `test*`), this solves that issue


McHaillet:
Do the unittests pass now on a device in exclusive process mode?

McHaillet (approved):

sroet:
> Do the unittests pass now on a device in exclusive process mode?

Yes! (and fail on `HEAD~1`)

PR #81, opened 2024-01-24
calculate spectrum whitening filter only in the search region

McHaillet:
Small patch to only calculate the whitening filter in the user specified search region. Edges of reconstructions can often have reconstruction/interpolation artifacts that users might want to avoid. The whitening filter should also not take them into account as the data here does not contain reliable info about the signal to noise ratio of the tomogram.


sroet:
could you add a test? (probably a quick one that checks if the saved whitening filter has the expected dimensions?) LGTM otherwise

McHaillet:
Test is added!

I also switched to recalculating the whitening filter in TMJob's. Since version 0.3.3 its fast to calculate and this will prevent
- bugs later on where the whitening filter does not have the correct size for a cropped search volume
- and properly update the whitening filter to the search region

sroet (approved):
LGTM, feel free to merge

PR #82, opened 2024-01-26
Standard deviation cut-off lower threshold

McHaillet:
Took longer than expected. 
Some observations:
- I tried taking the sqrt before the cutoff but this failed as apparently negative values can be returned from line 201. Maybe has to do with fourier transform of a square box, which has to be approximated with infinite waves in fourier space?
- Tried setting to lower limit, but this resulted in `test_template_matching.py` finding the object back in a completely wrong position.
- Leaving it at 1 but lowering the limit works, all tests pass.

As an alternative solution, I could also normalize the search volume in each process before going to the GPU template matching structure, by doing something like:
search_volume = (search_volume - search_volume.mean())/ search_volume.std()


sroet (approved):
LGTM, feel free to merge

PR #83, opened 2024-01-30
Subvolume indexing bug

McHaillet:
Noticed this bug this morning when I was calculating something myself. 

Apparently the subvolumes were not placed back correctly into the large subvolume when merging subjobs. This was because the whole_start parameter was not correctly calculated as I was subtracting the search_origin in x from both the x,y and z dimension. 

I first wrote a test to highlight the bug and after updating it the bug is now indeed fixed.


sroet (approved):
LGTM, nice test driven development :wink: . Feel free to merge

PR #87, opened 2024-02-02
Add scipy to requirements in setup.py

### LINKED ISSUE #85: add scipy to package dependencies
Just noticed its not explicitly listed as a dependency but implicitly installed because of cupy. Might be better to add it?

sroet:
closes #85


McHaillet:
Thanks!

PR #90, opened 2024-02-06
make json writing OS agnostic in tmjob.py

### LINKED ISSUE #89: Make json dumping OS agnostic
https://github.com/SBC-Utrecht/pytom-template-matching-gpu/blob/7d60c5be3dcf868892d6d9268f609ae3472ea42a/src/pytom_tm/tmjob.py#L219

This line does not work on Windows, we should change it to the parent class instead

sroet:
I `grep`ed through the codebase and this was the only instance of `PosixPath` that came back.
This is untestable as `pathlib.WindowsPath('.')` will error out with a `NotImplementedError` on unix systems

closes #89


sroet:
tests still pass :smile:

PR #98, opened 2024-02-09
Template.py: make filter warnings conditional and allow for no ctf_params

sroet:
replaces #88

I got a question from a novice user that thought it was bad when he did not specify the lowpass filter.
This now makes the log message conditional to the input

During the writing of the test I saw that there was code handeling `ctf_params=None` (line 73), so I added that option as a default to the call as well

### DIFF
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -20,7 +20,7 @@ def generate_template_from_map(
         input_map: npt.NDArray[float],
         input_spacing: float,
         output_spacing: float,
-        ctf_params: dict,
+        ctf_params: Optional[dict] = None,
         center: bool = False,
         filter_to_resolution: Optional[float] = None,
         output_box_size: Optional[int] = None,
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -37,8 +37,9 @@ def generate_template_from_map(
         )
 
     if filter_to_resolution is None or filter_to_resolution < (2 * output_spacing):
-        logging.warning(f'Filter resolution is either not specified (in which case you can ignore this warning) '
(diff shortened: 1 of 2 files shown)


McHaillet:
Thanks for the PR to address it!

I was thinking it might be better if there is no warning message if the filter is not set.
Only if its set and too low, it should give a warning that it is resseting it to nyquist resolution.
```
if filter is None:

**Comment by sroet** [comment]
> Thanks for the PR to address it!
> 
> I was thinking it might be better if there is no warning message if the filter is not set. Only if its set and too low, it should give a warning that it is reseting to nyquist resolution. Something like:
> 
> ```
> if filter_to_resolution is None:
>     filter_to_resolution = 2 * output_spacing
> elif filter_to_resolution < (2 * output_spacing):
>     logging.warning('...')
>     filter_to_resolution = 2 * output_spacing
> ```

You are right, was my preferred solution as well 👍 I altered it to that, please have another look

File: src/pytom_tm/template.py [hunk 1]
```diff
@@ -37,8 +37,12 @@ def generate_template_from_map(
         )
 
     if filter_to_resolution is None or filter_to_resolution < (2 * output_spacing):
```
McHaillet, line 39:
```suggestion
    if filter_to_resolution is None:
```
Like this right?

McHaillet (changes requested):

src/pytom_tm/template.py
sroet: yeah, forgot to actually pull int the latest code on my test machine...

sroet: the test do fail on commit e4482cf, as they should xD

McHaillet: haha good

McHaillet (approved):
LGTM

Issue #106, opened 2024-02-14
cupy v13 breaks ReductionKernel


McHaillet:
I found that the update to cupy v13 is actually breaking for the package. It looks also like the same issue that was reported here (issue #80) at first.

Installing on the cluster with:

```
conda create ... python=3
python -m pip install cupy=12.3.0
python -m pip install .[plotting]
```

I run the tests without issue:
```
Ran 22 tests in 24.726s

OK
```

However with:

```
conda create ... python=3
python -m pip install cupy=13.0.0
python -m pip install .[plotting]
```

The tests break:
```
Ran 15 tests in 274.780s

FAILED (errors=3)
```

Specifically I get this error:
```
Traceback (most recent call last):
  File "/data2/mchaillet/programs/tm_gpu_test/pytom-template-matching-gpu/tests/test_tmjob.py", line 85, in setUpClass
    score, angle = job.start_job(0, return_volumes=True)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data2/mchaillet/programs/tm_gpu_test/env/lib/python3.12/site-packages/pytom_tm/tmjob.py", line 486, in start_job
    results = tm.run()
              ^^^^^^^^
  File "/data2/mchaillet/programs/tm_gpu_test/env/lib/python3.12/site-packages/pytom_tm/matching.py", line 176, in run
    self.stats['variance'] += (square_sum_kernel(self.plan.ccc_map) / self.plan.volume.size)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cupy/_core/_reduction.pyx", line 828, in cupy._core._reduction.ReductionKernel.__call__
  File "cupy/_core/_reduction.pyx", line 370, in cupy._core._reduction._AbstractReductionKernel._call
  File "cupy/_core/_cub_reduction.pyx", line 689, in cupy._core._cub_reduction._try_to_call_cub_reduction
  File "cupy/_core/_cub_reduction.pyx", line 526, in cupy._core._cub_reduction._launch_cub
  File "cupy/_core/_cub_reduction.pyx", line 461, in cupy._core._cub_reduction._cub_two_pass_launch
  File "cupy/_util.pyx", line 64, in cupy._util.memoize.decorator.ret
  File "cupy/_core/_cub_reduction.pyx", line 240, in cupy._core._cub_reduction._SimpleCubReductionKernel_get_cached_function
[...]

PR #107, opened 2024-02-14
Cupy13 workaround

McHaillet:
Temporary workaround for ReductionKernel bug in cupy 13. Code checks if cupy version 13 is installed and switch to simple 

```
def square_sum_kernel(x):
        return (x ** 2).sum()
```
, instead of custom kernel.

See issue #106 for more info. The custom kernel is still preferred as it provides a ~2 x speed up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">= 3.9"
 dependencies = [
     "numpy",
-    #"cupy",
+    "cupy",
     "voltools",
     "tqdm",
     "mrcfile",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-template-matching-gpu"
-version = "0.3.4"
+version = "0.3.5"
```
sroet, line 7: we never did a 0.3.4 release, so this doesn't have to change
```suggestion
version = "0.3.4"
```

sroet (changes requested):
version doesn't need to increase, but this looks good otherwise.

Did you test it?

pyproject.toml
McHaillet: ah okay

McHaillet:
Yes, tested it with cupy 12.3 and cupy 13, both work 👍

Issue #101, opened 2024-02-12
Processing speed and memory


rezaparaanczi:
Hello,

Thanks for the nice tutorial. I have been testing this on my data and it works great for ribosomes with a 40 Angstrom filter on tomograms with 21.12 A/pix pixel size (K3 camera, bin 8). I'm looking at another target and decided to do higher resolution template matching with the template at 22 Angstrom low pass filter, and tomogram pixel size of 10.56 A/pix (bin 4), sampling is 3 degrees. With these new setting I can only run pytom on gpus with 80GB of memory, as it's taking 42GB of memory during the run which is slightly more than the next grade gpus with 40GB of memory. I was wondering if I'm missing something, because according to the tutorial a 1080 Ti gpu has enough memory for this job and that only has ~10 GB of memory. That's the memory issue I wanted to raise. I also wanted to raise a speed issue, because with 4 A100 gpus it takes 28 hours for one tomogram. I would appreciate any suggestions. Here is an example command I'm running:
pytom_match_template.py -t templates/template.mrc -m templates/mask_10.56Apix.mrc -v paceview12_ts_001.mrc.2000.wbp.yz.tapered.rec -d results_RP_3deg_weightwhite2/ -a rawtlt/paceview12_ts_001.mrc.rawtlt --angular-search 3.00 --voxel-size 10.56 --low-pass 22 -g 0 1 2 3 --per-tilt-weighting --spectral-whitening

McHaillet:
Hi! Nice to hear its working!

The memory issue can be easily solved by splitting your tomogram into smaller patches, see the `-s`, `--volume-split` option for [pytom_match_template.py ](https://github.com/SBC-Utrecht/pytom-template-matching-gpu/wiki/template-matching). With the memory use you are indicating `-s 2 1 1` would already be sufficient, this would split the tomo once along the x-axis. If that is still too large, you could try `-s 3 1 1` or `-s 2 2 1`. 

> I also wanted to raise a speed issue, because with 4 A100 gpus it takes 28 hours for one tomogram.

That seems about right for such a volume/angular-search combination. Although on our cluster a tomo of size 928x960x500 with 3.00 search on 4 GPU's takes ~10h. I guess because of the K3 your tomos are slightly larger, but that does seem a factor 2 slower. In general though, the exhaustive search does take a long time on such dimensions. 

Some suggestions:
[...]

PR #108, opened 2024-02-14
update install instructions to prefer building cupy

McHaillet:
I updated the installation instuctions in the README to prefer letting cupy build against a preinstalled cuda toolkit as the gain in performance is almost two-fold on our cluster. 

I tried building cupy against a cuda toolkit in the miniconda environment, but the result was that cupy still installed against the system installation. So I just left it as the reliable option to pull it directly from conda-forge.

related to issue #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,18 +12,17 @@ nvidia-cuda-toolkit
 
 ## Installation
 
-There are 2 options for creating a conda environment. We recommend option (1) which builds cupy against a system 
-installed cuda-toolkit. Compared to option (2) this can give an almost two-fold speedup.
+There are 2 options for creating a conda environment. We recommend option (1) which will later allow cupy to build 
+against a system installed cuda-toolkit. Compared to option (2) this can give an almost two-fold speedup:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,35 +12,40 @@ nvidia-cuda-toolkit
...
-1. **(recommended)** Create a new environment with a prebuild cupy version and complete CUDA-toolkit. This is reliable but takes more disk space.
+1. **(recommended)** Create a new python 3 environment:
 
     ```commandline
-    conda create -n pytom_tm -c conda-forge python=3 cupy cuda-version=11.8
+    conda create -n pytom_tm python=3
     ```
 
-2. Create a new environment without cupy and let pip build against a system installed CUDA-toolkit.
+2.  Create a new environment with a prebuild cupy version and complete CUDA-toolkit. This is reliable, but takes more
```
sroet, line 24:
```suggestion
2.  Create a new environment with a prebuild cupy version and complete CUDA-toolkit. This is more reliable, but takes more 
```

sroet (approved):
LGTM, one small possible correction, but feel free to ignore

PR #109, opened 2024-02-15
Convert bin scripts to entry points

### LINKED ISSUE #5: switch to entry points instead of bin scripts

sroet:
closes #5, also allows for direct calling these functions with a given `argv` which should help testing

TODO:

- [X] make entry points file
- [x] check for double imports and move them to the top
- [x] move actual plotting code to its own file instead of bin script
- [x] format the new files with `ruff` and `black`
- [x] update pypoject.toml to install the new entry points as bin scripts
- [x] remove old `bin` directory
- [x] add tests that these scripts are installed and on the PATH
- [x] add tests that these entry points can be called with a `-h` flag
- [x] make sure code is still importable if matplotlib is not installed


sroet:
@McHaillet should this lead to an increase in version number? (this does technically add a bunch of functionality due to the `plotting.py` file)

src/pytom_tm/entry_points.py
```diff
@@ -0,0 +1,913 @@
...
+    write_mrc(
+        output_path,
+        np.flip(template, axis=0) if args.mirror else template,
+        args.output_voxel_size_angstrom,
+    )
+
+
+def estimate_roc(argv=None):
+    argv = _parse_argv(argv)
+    # TODO: actually move the code to this location
```
McHaillet, line 351:
```suggestion
```

McHaillet (approved):
LGTM, just noticed this line that looks like it can be removed now. A version bump also seems like a good idea.

PR #110, opened 2024-02-19
Slicing for std calculation

### LINKED ISSUE #97: pass subvolume slicing to TemplateMatchingGPU for improved std
Currently for tomograms that are tiled into patches for GPU memory constraints are used fully to calculate the standard deviation of the search. However, there always needs to be overhang between patches (namely template_size // 2) to get an accurate score map for the whole tomo. This results in standard deviations being averaged over areas that are calculated multiple times, leading to a incorrect final search standard deviation.

It could be easily fixed by passing a slice to the template matching run based on indexes that are already present to place the tile back into the full tomo, and only calculate the std in the sliced area.

McHaillet:
This PR improves the calculation of the standard deviation over a template matching search. It closes #97.

First some background on why I updated this. Template matching has a huge search space (N_voxels * rotations) which is mainly false positives, and has in comparison a tiny fraction of true positives. If we have a Gaussian for the background (with expected mean 0 and some standard deviation), the false alarm rate can be calculated for a certain cut-off value, as it is dependent on the size of the search space. For example, a false alarm rate of (N_voxels * rotations)^(-1), indicates it would expect 1 false positive in the whole search. This can be calculated with error function,

$$N^{-1} = erfc( \theta / ( \sigma \sqrt{2} ) ) / 2$$

, where theta is the cut-off, sigma the standard deviation of the Gaussian, and N the search space.

The search space is easily calculated, the standard deviation can be kept tracked of for each orientation (which is what this custom square_sum_kernel is used for). However there were true problems:
- with the next_fast_len for ffts the volume was padded with zeros that were also incorporated in the square_sum_kernel.
- with subvolume splitting I added overhang between subvolumes, needed for accurate scores, but these regions were incorporated doubly in the standard deviation calculation.

I fixed it by passing a search_volume_roi to template matching which contains the actual region of interest wihout fft padding and without template overhang. The indexing is already calculated, so I just created a slicing for the cupy array.

I first added tests to check whether the search_space and std are consistent with subvolume splitting (and also rotation splitting), and realised there was a bug in tmjob.py in calculating the start of the subvolume (lines 380 to 384): I did not put brackets around `template_size // 2` which messes with integer division with the minus sign in front :weary: . This is now fixed. 

I also added plotting of an extraction graph which shows a histogram of extracted scores together with the background Gaussian (as this is now properly estimated). Which I think is nice for users to see. 

Let me know if anything is unclear.

### DIFF
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -129,9 +129,9 @@ class TemplateMatchingGPU:
         self.stats = {'search_space': 0, 'variance': 0., 'std': 0.}
         if stats_roi is None:
             self.stats_roi = (
-                slice(0, volume.shape[0]),
-                slice(0, volume.shape[1]),
-                slice(0, volume.shape[2])
+                slice(None),
+                slice(None),
+                slice(None)
             )
         else:
             self.stats_roi = stats_roi
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -243,9 +243,9 @@ class TemplateMatchingGPU:
             )
 
             self.stats['variance'] += (
-                    square_sum_kernel(
-                        self.plan.ccc_map[self.stats_roi]
-                    ) / roi_size
+                square_sum_kernel(
+                    self.plan.ccc_map[self.stats_roi]
+                ) / roi_size
             )
 
         self.stats['search_space'] = int(roi_size * len(self.angle_ids))
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -86,6 +86,7 @@ class TemplateMatchingGPU:
             mask: npt.NDArray[float],
             angle_list: list[tuple[float, float, float]],
             angle_ids: list[int],
+            stats_roi: tuple[slice, slice, slice],
             mask_is_spherical: bool = True,
             wedge: Optional[npt.NDArray[float]] = None
     ):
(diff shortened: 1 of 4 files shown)


src/pytom_tm/matching.py
```diff
@@ -123,6 +127,14 @@ def __init__(
         self.angle_list = angle_list
         self.angle_ids = angle_ids
         self.stats = {'search_space': 0, 'variance': 0., 'std': 0.}
+        if stats_roi is None:
+            self.stats_roi = (
+                slice(0, volume.shape[0]),
+                slice(0, volume.shape[1]),
+                slice(0, volume.shape[2])
```
sroet, line 134: if the intention is to slice nothing, then `slice(None)` is probably better
```suggestion
                slice(None),
                slice(None),
                slice(None)
```

tests/test_tmjob.py
```diff
@@ -29,6 +29,8 @@
 TEST_ANGLES = TEST_DATA_DIR.joinpath('tomogram_angles.mrc')
 TEST_CUSTOM_ANGULAR_SEARCH = TEST_DATA_DIR.joinpath('custom_angular_search.txt')
 TEST_WHITENING_FILTER = TEST_DATA_DIR.joinpath('tomogram_whitening_filter.npy')
+TEST_JOB_JSON = TEST_DATA_DIR.joinpath('tomogram_job.json')
+TEST_JOB_JSON = TEST_DATA_DIR.joinpath('tomogram_job.json')
```
sroet, line 33: double, 1 should be removed
```suggestion
TEST_JOB_JSON = TEST_DATA_DIR.joinpath('tomogram_job.json')
```

sroet (approved):
1 suggestion and 1 line was added twice, LGTM otherwise

sroet (changes requested):
(selected the wrong option)

src/pytom_tm/matching.py
McHaillet: good one, that's easier to read as well

sroet (approved):
LGTM, feel free to merge

PR #112, opened 2024-02-19
Prevent whitening for extraction

### LINKED ISSUE #95: whitening filter should not be calculated for extraction job
When I lately set the whitening filter to recalculate when a template matching run is started, I forgot this also means its recalculated when a .json job file is reloaded for the particle annotation script. I guess when a stored .json job is loaded into the TMJob class it should not recalculate the whitening filter.

McHaillet:
This closes #95. Whitening filter was recalculated when loading a job for particle assignment. An unnecessary step that can take quite a while for large tomograms. I added a flag to load_json_to_tmjob and the TMJob class init that can be set for particle assignment and turns off recalculation. Also added some tests around it.

### DIFF
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -16,13 +16,16 @@ from pytom_tm.weights import create_wedge, power_spectrum_profile, profile_to_we
 from pytom_tm.io import read_mrc_meta_data, read_mrc, write_mrc, UnequalSpacingError
 
 
-def load_json_to_tmjob(file_name: pathlib.Path) -> TMJob:
+def load_json_to_tmjob(file_name: pathlib.Path, load_for_extraction: bool = True) -> TMJob:
     """Load a previous job that was stored with TMJob.write_to_json().
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -16,13 +16,16 @@
 from pytom_tm.io import read_mrc_meta_data, read_mrc, write_mrc, UnequalSpacingError
 
 
-def load_json_to_tmjob(file_name: pathlib.Path) -> TMJob:
+def load_json_to_tmjob(file_name: pathlib.Path, load_for_extraction: bool = True) -> TMJob:
```
sroet, line 19: With this and the previous PR, we should also probably update the version number on our pyproject.toml

sroet (approved):
LGTM, one comment about versioning, but feel free to leave that for another PR

src/pytom_tm/tmjob.py
McHaillet: Good point, will update !

sroet (approved):
LGTM, feel free to merge

PR #114, opened 2024-02-21
small fixes for template generation and preparation

McHaillet:
Two small changes here because I encountered issues with these things in the dataset I am processing.

Template generation used to do padding with edge mode (left image) and now instead pads with zeros (right). This reduces edge artifacts if the reference structure has a zeroed background (which should be the case for EM maps!).

pytom-template_with_and_without_zero_padding

Second change is that the template is now always multiplied with the mask before starting template matching (instead of after applying the wedge), this significantly reduces artifacts from convolution with the missing wedge. Left image is before, right image is after:

pytom-wedge_with_and_without_prior_masking


sroet (approved):
LGTM! feel free to merge

PR #116, opened 2024-02-21
get gh workflows working

sroet:
Let's see if this gets run before merging 🤞

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -12,14 +12,12 @@ on:
     # Weekly on saturday 23:59
     - cron: "59 23 * * 6"
 
-defaults:
-  runs-on: [self-hosted, linux, x64, gpu]
-  container:
-    image: continuumio/miniconda3
-    options: --runtime=nvidia --gpus all
-
 jobs:
   test_suite:
+    runs-on: [self-hosted, linux, x64, gpu]
+    container:
+      image: continuumio/miniconda3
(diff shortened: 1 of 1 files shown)


sroet:
w00t it works (and gets the gpus), lets see if I can get the tests to run (and pass)

sroet:
It is successfully running the tests! @McHaillet please have a look 😄

McHaillet:
Wow, very nice! So this also runs for PRs? And then once a week for the main branch?

McHaillet:
Ah wait, I see it in the options. So for: PRs to main, pushes to main, and weekly.

Would this also mean it triggers if we push to main in our forks?

sroet:
> Wow, very nice! So this also runs for PRs? And then once a week for the main branch?

It runs on PRs to main (should require approval from one of us for people outside of SBC), push of a tag release tag to main, and scheduled on the default branch (main in this case)
```
on:
  pull_request:
    branches:
      - main
  push:
    branches:
      - main
    tags:
      - "[0-9]+.[0-9]+.[0-9]+"
  schedule:
    # Weekly on saturday 23:59
    - cron: "59 23 * * 6"

sroet:
> Would this also mean it triggers if we push to main in our forks?

It does and will fail as our own forks don't have access to the self-hosted runner. I would advise you to disable actions in your own fork

McHaillet (approved):
LGTM!

PR #118, opened 2024-02-21
add coverage reporting

sroet:
This should add coverage and coverage reporting as a github action

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -28,8 +28,12 @@ jobs:
           name: coverage
           github-token: ${{ secrets.GITHUB_TOKEN }}
           run-id: ${{ github.event.workflow_run.id}}
+      - name: get PR
+        id: get-pr
+        run: echo "PR=$(cat coverage.PR)" >> "$GITHUB_OUTPUT"
       - name: Get Coverage
-        uses: orgoro/coverage@v3.1
+        uses: 5monkeys/cobertura-action@master
         with:
-          coverageFile: coverage.xml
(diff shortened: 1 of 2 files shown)


sroet:
hmm this can't easily be done with github workflows, without external parties due to github security. (It does not allow any workflows that run on a PR from forks on public repos to have write access to anything, which is required to write a comment on the PR with the required info).

There is technically an example roundabout way of dealing with it: https://github.com/py-cov-action/python-coverage-comment-action which I could look into.

Otherwise, I can probably set it up with an external party: 
- [codecov.io](https://about.codecov.io/) ([an example PR](https://github.com/dwhswenson/contact_map/pull/133), click on "view details" on the merge commit to see the actions/checks on top of the comment)
- [coveralls.io](https://coveralls.io/) ([example PR](https://github.com/infretis/infretis/pull/77), does not comment, but does give a check)
- [codacy](https://www.codacy.com/) (I don't think I ever set up this one, but do know it exist. [example PR](https://github.com/codacy/codacy-coverage-reporter/pull/478))
- Any other?

@McHaillet which of the following would you prefer? (I slightly prefer to see if I can get the roundabout way to work)

sroet:
Hmm, I just realized that the codecov reporting code will probably not work as it pulls some data to figure out which PR to post stuff to, which is then run on a pipline that has no reference to that PR...

sroet:
alright, switched to another package that should work
@McHaillet please have a look (it will not work properly before `coverage.yml` is actually in the repo)

PR #121, opened 2024-02-22
Update unit-tests.yml

sroet:
Actually run coverage on the code (requires editable install)


sroet:
@McHaillet This is what the report would look like if it actually ran. Good enough for you?

McHaillet (approved):
LGTM

PR #127, opened 2024-02-27
add tests for missing libraries

sroet:
This should add coverage for the lines that deal with missing optional modules


sroet:
@McHaillet please wait on the review,  python import caching is anoying for these tests

McHaillet (changes requested):
Looks good!

I guess the `import matplotlib` line in extract.py can be removed (dont know why I put that there in the first place): https://github.com/sroet/pytom-template-matching-gpu/blob/4418d20aec37942f55785406a7dc893784d3c3d8/src/pytom_tm/extract.py#L20

McHaillet (approved):
LGTM!

PR #133, opened 2024-02-29
Update main branch version to 0.4.0

sroet:
The release of last week 0.3.5 was done on a version of the code that did not have the correct version number. So when I updated it in #109 it was still one too few.

I am also open to changing this to 0.4.0 because of the package renaming
[EDIT}: after an offline discussion we decided to go for 0.4.0 because of the name change


sroet:
also after this we should cut a release to enable the zenodo doi tracking

McHaillet (approved):
LGTM

PR #137, opened 2024-03-01
Fix release workflow

sroet:
this commands succesfully pulls the release from testpypi locally


McHaillet (approved):
Good, wrong link or something?

sroet:
> Good, wrong link or something?
yeah the `/legacy/` is for uploads, but the `/simple/` is for downloads if I follow the current documentation

PR #138, opened 2024-03-01
Try to fix race condition in testing

sroet:
Tests had a non-insignificant chance of failing due to this condition being triggerd.
I added a 1 sec wait between raising the error and seeing if all children are dead to hopefully solve this issue


McHaillet (approved):
LGTM! I assume it fixes it, but if not we'll find out.

PR #139, opened 2024-03-01
update readme to point to pip install instead of source

sroet:
Now that the pip install commands are working, we should update our installation instructions as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,13 @@ python -m pip install pytom-match-pick
 ```
 
 ## Cupy warning
-Trying to import the code with 
+Having issues running the software? If cupy is not correctly installed, 
 ```commandline
 python -c "import pytom_tm"
 ```
 
 can show a cupy warning. If this is the case, this probably means cupy is not correctly installed.
-If your cupy installation is broken, the following should raise some errors:
+Alternatively, cupy can sometimes be installed without issue but not detect CUDA correctly. In that case, the following should raise some errors:
 ```commandline
 python -c "import cupy as cp; a = cp.zeros((100,100))"
 ```
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ Once the environment is created, activate it:
 conda activate pytom_tm
 ```
 
-The install the code with `pip` (building cupy can take a while!):
+Then install the code with `pip` (building cupy can take a while!):
 
 ```commandline
-python -m pip install pytom-match-pick.[plotting]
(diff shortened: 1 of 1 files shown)


sroet:
The tests fail with the same issue that #138 tries to fix, so this shouldprobably be merged after that one

README.md
```diff
@@ -35,33 +35,28 @@ Once the environment is created, activate it:
...
-python -m pip install .
+python -m pip install pytom-match-pick
 ```
 
-## Tests
-
-To run the unittests:
-
+## Cupy warning
+Trying to import the code with
```
McHaillet, line 53:
```suggestion
Having issues running the software? If cupy is not correctly installed, 
```

README.md
```diff
@@ -35,33 +35,28 @@ Once the environment is created, activate it:
...
+Trying to import the code with 
 ```commandline
-cd tests
-python -m unittest discover
+python -c "import pytom_tm"
 ```
 
-If cupy is not correctly installed, this should raise some errors:
-
+can show a cupy warning. If this is the case, this probably means cupy is not correctly installed.
```
McHaillet, line 58:
```suggestion
will show an import error. If this is the case, this probably means cupy is not correctly installed.

```

README.md
```diff
@@ -35,33 +35,28 @@ Once the environment is created, activate it:
...
+If your cupy installation is broken, the following should raise some errors:
```
McHaillet, line 59:
```suggestion
Alternatively, cupy can sometimes be installed without issue but not detect CUDA correctly. In that case, the following should raise some errors:
```

McHaillet (changes requested):
Installation part is good!

For the cupy errors, some suggested changes and I also think this part,

> To solve cupy installation issues, please check [the cupy docs](https://docs.cupy.dev/en/stable/install.html#installing-cupy). It might be solved by installing a specific build compatible with the installed cuda toolkit. 

, could be changed to first trying to install with the conda cuda-toolkit and cupy option. And if that doesnt work looking at the cupy docs.

README.md
sroet: Currently the code only does logging.warning, not an actually import error

sroet:
@McHaillet, one of your suggestions isn't accurate with the current code behavior. I have implemented the others

README.md
McHaillet: Fair!

McHaillet (approved):
LGTM!

PR #140, opened 2024-03-05
change default mrc permissive to false, try to autorecover, remove userfacing 'permissive' argument

### LINKED ISSUE #39: MRC reading permissiveness needs to be strict and tested
Reading MRC files with permissive mode needs the following things:
- unittests to determine what mrc headers are allowed
- user message to inform about mrcfile not sticking to the format

 > LGTM, I would like to also have an added test so that we can read slightly borked meta_data but still give a reasonable error if it is completely useless. But for a quick fix this seems fine

_Originally posted by @sroet in https://github.com/SBC-Utrecht/pytom-template-matching-gpu/pull/38#pullrequestreview-1612981246_

sroet:
closes #39 

Done:

- [x] make a test with an unrecoverable mrc file
- [x] remove `permissive` from `read_mrc` and `read_mrc_meta_data`
- [x] update version for API break


sroet:
@McHaillet  Did you have some example failing MRCs as were mentioned in #35 ?

sroet:
@McHaillet It took a while, but this is ready for a round of review 😄 Feel free to throw some other files at it to see if you like/dislike the behavior

McHaillet (approved):
Looks very solid too me, really nothing to too comments 👍 

Don't really have any more examples of corrupt mrcs, so nothing to further test from my side. Feel free to merge!

PR #141, opened 2024-03-06
ctf model ramp weighting correction and updated interpolation center

McHaillet:
This PR updates internal code in _create_tilt_weighted_wedge to create a more precise weighting function.

First I add a linear weighting up to the point where tilt image overlap in Fourier space. This now allows summation of the tilts instead of doing a maximum operation. It reduces aliasing artifacts quite a lot:

Before:
ctf_model_before_0 5

After:
ctf_model_after_0 5

The second update (which is also visible in the image, but difficult to see), is that I changed the interpolation center when tilting the images. I switched to voltools instead of ndimage.rotate as this allows me to set the center to the origin in Fourier space, while ndimage.rotate instead uses (shape - 1) / 2, which is not correct for fourier space. In the above image the wedge is shifted one pixel to the left. But this can be quite crucial for the overlap of low frequencies between the tomo and template.

### DIFF
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -513,22 +513,16 @@ def _create_tilt_weighted_wedge(
                     ctf_params_per_tilt[i]['cs'] * 1e-3,
                     flip_phase=True  # creating a per tilt ctf is hard if the phase is not flipped
                 ), axes=0,
-            ) * exact_weighting
+            )
             tilt[:, :, image_size // 2] = np.concatenate(
                 (  # duplicate and flip the CTF around the 0 frequency; then concatenate to make it non-reduced
                     np.flip(ctf[:, 1: 1 + image_size - ctf.shape[1]], axis=1),
                     ctf
                 ),
                 axis=1
-            )
+            ) * exact_weighting
         else:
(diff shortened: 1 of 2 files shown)


McHaillet:
And I updated the version to 0.5.0 to cut a release with this update!

src/pytom_tm/weights.py
```diff
@@ -536,7 +555,8 @@ def _create_tilt_weighted_wedge(
                     rotated *
                     np.cos(alpha)  # apply tilt-dependent weighting
             )
-        tilt_weighted_wedge = np.maximum(tilt_weighted_wedge, weighted_tilt)
+
+        tilt_weighted_wedge += weighted_tilt  # np.maximum(tilt_weighted_wedge, weighted_tilt)
```
sroet, line 559: leftover old code
```suggestion
        tilt_weighted_wedge += weighted_tilt
```

sroet (changes requested):
one leftover old code LGTM otherwise

src/pytom_tm/weights.py
McHaillet: Thanks for spotting :)

sroet (approved):
LGTM feel free to merge

PR #147, opened 2024-03-26
don't make plots for particle extraction in estimate_roc

### LINKED ISSUE #143: estimate_roc should call extraction with plotting false flag
As per title, the estimate_roc entry point should call `extract_particles()` with the `create_plot=False` flag!

sroet:
closes #143


McHaillet (approved):
Thanks for the fix!

PR #149, opened 2024-03-26
change 'ptm' to 'rln' in the extraction star file

### LINKED ISSUE #93: switch star file headers to _rln instead of _ptm
Many software fails will the _ptm headers (although that shouldnt be the case), alternatively could also go to the teamtomo/cryohub package and add it as a possibility (which makes reading in Blik easier).

sroet:
should close #93

@McHaillet I did not see any other reference to 'ptm', however these do not have the underscore, is that correct?

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -260,17 +260,17 @@ def extract_particles(
         ] *= cut_mask
 
     output = pd.DataFrame(data, columns=[
-        'ptmCoordinateX',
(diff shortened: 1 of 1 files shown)


McHaillet:
> @McHaillet I did not see any other reference to 'ptm', however these do not have the underscore, is that correct?

Yep, that's right. starfile adds the underscore before the column headers!

McHaillet (approved):
LGTM

PR #157, opened 2024-04-04
Phase plate ctf and argparse grouping

### LINKED ISSUE #76: phase shift in CTF to model vpp phase plate
Hi, 
I collected some vpp phase plate tomography datasets and would like to try pytom-template matching. I was wondering how to set parameters when using pytom_create_template.py. Is it different from conventional defocus tomography dataset? Do I need to consider the phase shift?  
Best,
Peter

### LINKED ISSUE #150: argparse grouping for readability
The pytom_match_template entry point could benefit from argument grouping to improve the readability for users. For example a group for CTF input parameters and a group for volume controls (splitting, search range restriction), could already do a lot!

see here: https://docs.python.org/3.7/library/argparse.html#argparse.ArgumentParser.add_argument_group

McHaillet:
In this PR I added the following:

- grouping of argparse parameters for pytom_match_template.py to make the help more readable!
- a phase shift parameter for CTF curves, with allows users to model a phase plate. Phase shift calculation follows standard conventions. 

For the phase shift parameter I had to make job reading backwards compatible, so I added another version based check.

Closes #150 and closes #76


src/pytom_tm/entry_points.py
```diff
@@ -568,38 +576,24 @@ def match_template(argv=None):
     parser = argparse.ArgumentParser(
         description="Run template matching. -- Marten Chaillet (@McHaillet)"
     )
-    parser.add_argument(
+    io_group = parser.add_argument_group("Template, search volume and output")
```
sroet, line 579: oxford comma reads nicer in my opinion
```suggestion
    io_group = parser.add_argument_group("Template, search volume, and output")
```

src/pytom_tm/tmjob.py
```diff
@@ -61,6 +61,13 @@ def load_json_to_tmjob(file_name: pathlib.Path, load_for_extraction: bool = True
         pytom_tm_version_number=data.get('pytom_tm_version_number', '0.3.0'),
         job_loaded_for_extraction=load_for_extraction,
     )
+    # if the file originates from an old version set the phase shift for compatibility
+    if (
+        version.parse(job.pytom_tm_version_number) < version.parse('0.6.1') and
+        job.ctf_data is not None
+    ):
+        for tilt in job.ctf_data:
+            tilt['phase_shift_deg'] = .0
```
sroet, line 70: This is not hit in the tests, can you add a test to make sure this compatibility check works as expected?

src/pytom_tm/weights.py
```diff
@@ -609,7 +617,7 @@ def create_ctf(
     tan_term = np.arctan(amplitude_contrast / np.sqrt(1 - amplitude_contrast ** 2))
 
     # determine the ctf
-    ctf = - np.sin(chi + tan_term)
+    ctf = - np.sin(chi + tan_term - np.deg2rad(phase_shift_deg))
```
sroet, line 620: the reference for the convention (https://github.com/dtegunov/tom_deconv) uses `cos` instead of `sin` does that mater for the "0-point" of the phase shift?

src/pytom_tm/weights.py
```diff
@@ -6,6 +6,8 @@
 from typing import Optional, Union
 from pytom_tm.io import UnequalSpacingError
 from itertools import pairwise
+from pytom_tm import __version__ as PYTOM_TM_VERSION
+from packaging import version
```
sroet, line 10: Not actually used in this file, I think
```suggestion
```

sroet (changes requested):
1 missing test, 2 leftover imports, 1 style comment (feel free to ignore) and a question about the convention. LGTM otherwise

src/pytom_tm/weights.py
McHaillet: very good spot, thanks

really need to double check this!

McHaillet: You were absolutely right. Below figures illustrating, comparison is a bit weird as the CTF definition in this tom_deconv is exactly inverse. So without the phase shift and same parameters I get:

ctf_comparison_without_phase_plate

With the phase shift, I indeed need to add the term (instead of subtract) to get the 'correctly' inversed curves:

ctf_comparison_with_phase_plate

src/pytom_tm/tmjob.py
McHaillet: I just now added a job file from before. Had to change the file paths in the job to accommodate to point to the dynamically generated test files.

src/pytom_tm/weights.py
McHaillet: Let me know if it makes sense

sroet: > Let me know if it makes sense

Certainly does! thanks for checking :+1:

sroet (approved):
LGTM, thanks for fixing my comments.

Feel free to merge

PR #159, opened 2024-04-06
default value phase shift entry point

McHaillet:
While running realised this issue: phase shift needs to be 0 by default otherwise None gets passed to CTF creation.


sroet (approved):
LGTM, feel free to merge

McHaillet:
Thanks for the quick review ;) Was  expecting you would check on monday

PR #163, opened 2024-04-30
auto generate sampling angles

### LINKED ISSUE #92: angular sampling options
It would be good if there were some more options for choosing angular increments in searches. Especially for exhaustive searches: the most exhaustive is 3. degrees, while the second most exhaustive is 7. degrees. Here it would be important if users have availability for a 5. degrees increment searches.

Ideally they could be generated on the fly, but a smaller solution could already be to just include an added 5 degrees increment search.

sroet:
Will close #92 by auto generating `theta` and `phi` with [`healpix`](https://github.com/ntessore/healpix/tree/main) and sampling `psi` linearly. It will use the input angle as a `maximum` angle difference

Todo:

- [x] remove pre-generated angles
- [x] add angle autogeneration code
- [x] add new dependencies
- [x] make sure install still works
- [x] Update tests (except angle tests) to new type of code
- [x] update entry points to deal with new angle creation code
- [x] add angle generation tests
- [x] add any other tests to cover touched code (except entry_points)
- [X] test all added functions with mypy
- ~test all touched code with ruff~ (decided to leave this for another PR)
- ~format all touched files with black~ (decided to leave this for another PR)

# Other fixes:
The test of bad angles in the angle file was missing newlines, so it did raise the expected type `ValueError`, just the wrong `ValueError`

The coverage has increased to 81%


McHaillet:
Very excited to see this one!
We could also provide a feature to let users input the particle diameter, and calculatie the 'correct' sampling from that and  the pixel size (or low pass filter)

sroet:
> We could also provide a feature to let users input the particle diameter, and calculatie the 'correct' sampling from that and the pixel size (or low pass filter)

I agree, but think that is outside of the scope for the current PR. DId make issue #164 to keep track of it

sroet:
@McHaillet pulling in the code from #165 results in a race condition in the tests. 
I can't reproduce the error in the unit tests locally.
Do you know have any idea what could have introduced that? (either on that PR or on this one)

McHaillet:
for me locally they are running

> ```python
> FAIL: test_search_non_spherical_mask (test_template_matching.TestTM.test_search_non_spherical_mask)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/sander/github_files/pytom-template-matching-gpu/tests/test_template_matching.py", line 88, in test_search_non_spherical_mask
>     self.assertAlmostEqual(stats['std'], 0.005163, places=5,
> AssertionError: 0.0051728631983818125 != 0.005163 within 5 places (9.863198381812426e-06 difference) : Standard deviation of the search should be almost equal
> ```

The `places` argument should be 4 in the main branch. Might be the issue?

McHaillet:
In your branch it is still 5: https://github.com/sroet/pytom-template-matching-gpu/blob/6c46c2573661e7beedba83c9c114b12b32aab7eb/tests/test_template_matching.py#L88

Don't know why it didnt update, maybe try to rebase main?

sroet:
> In your branch it is still 5: https://github.com/sroet/pytom-template-matching-gpu/blob/6c46c2573661e7beedba83c9c114b12b32aab7eb/tests/test_template_matching.py#L88
> 
> Don't know why it didnt update, maybe try to rebase main?

I copied my version of the tests instead of using the one added in main and missed that it wasn't identical between the two tests

McHaillet:
comparison I ran on the tutorial data using this command with previous version (bench) and healpix sampling. This is the command:

```
pytom_match_template.py  -t templates/80S_ctf.mrc  -m templates/mask.mrc  -v dataset/tomo200528_100.mrc  -d results_80S_bench_healpix/  -a dataset/tomo200528_100.rawtlt  --angular-search 7.00  --voxel-size 13.79  -g 0  --low-pass 40
```

I then ran:

```
pytom_extract_candidates.py -j results_80S_bench_healpix/tomo200528_100_job.json -n 1000 -r 7
```

I loaded the resulting particle list and plot the rlnLCCmax column for both lists:

angular_search_lpf

Would have liked to see that the curve would have moved fully up, seeing that this increased the sampling slightly to 6.5 deg instead of 7 deg. (from 45000>50000 rotations). Now some particles got higher scores and others slightly lower. I guess the previous lists were slightly more equilibrated, something that healpix of course not fully guarantees.

The increased speed and dynamic generation is great though, so that in comparison to perhaps slight decrease in equilibration is probably worth it.

src/pytom_tm/angles.py
```diff
@@ -1,26 +1,48 @@
...
+    while hp.nside2npix(nside) < npix:
+        nside += 1
+    used_npix = hp.nside2npix(nside)
+    used_angle_diff = (4*np.pi/used_npix)**0.5 * (180/np.pi)
+    logging.info(f"Using an angle difference of {used_angle_diff:.4f} for Z1 and X")
+    theta, phi = hp.pix2ang(nside, np.arange(used_npix))
+    # Now for psi
+    n_psi_angles = int(np.ceil(360/angle_diff))
+    psi, used_psi_diff = np.linspace(0, 2*np.pi, n_psi_angles, endpoint=False, retstep=True)
+    logging.info(f"Using an angle difference of {used_psi_diff:.4f} for Z2")
```
McHaillet, line 41:
```suggestion
    logging.info(f"Using an angle difference of {np.rad2deg(used_psi_diff):.4f} for Z2")
```
Noticed when running that this should be converted to degrees!

src/pytom_tm/entry_points.py
```diff
@@ -637,8 +637,9 @@ def match_template(argv=None):
         "--angular-search",
         type=str,
         required=True,
-        help="Options are: [7.00, 35.76, 19.95, 90.00, 18.00, "
-             "12.85, 38.53, 11.00, 17.86, 25.25, 50.00, 3.00].\n"
+        help="If given a float it will generate an angle list with healpix for Z1 and X1 "
+             "and linear search for Z2. The given angle given will be used as the maximum for the "
```
McHaillet, line 641:
```suggestion
             "and linear search for Z2. The provided angle will be used as the maximum for the "
```

src/pytom_tm/tmjob.py
```diff
@@ -233,18 +235,33 @@ def __init__(
...
-        except KeyError:
+            angle_increment = float(angle_increment)
+            angle_is_float = True
+        except (ValueError, TypeError):
+            angle_is_float = False
+        if angle_is_float:
+            # Generate angle list on the fly
+            logging.info(f"Will generate an angle list with a maximum increment of {angle_increment}")
+            self.rotation_file = None
+            self.angle_list = angle_to_angle_list(angle_increment)
```
McHaillet, line 249: The list is now regenerated when loading a job file for extraction, fine by me as its fast (although the printing to terminal might be confusing). Potentially, could make use of the keyword `job_loaded_for_extraction` that I introduced for a similar issue to skip it?

McHaillet: The full angle_list is now also written to the json file. That seems a bit unnecessary and makes the json job file difficult to interpret. As the generation is anyway fast and reproducible, maybe we can skip saving it ?

Might be easier to generate the list in TMJob.start_job() ? That saves the extraction recalculation and prevents saving to file

McHaillet (changes requested):
Looks good, but I have a few suggestions.

A more general thing I noticed is that I could not load old job files, so not backwards compatible. Perhaps my suggestion of not storing the angle list to the json file would help (see below)?

I guess the issue is though that the jobs previously pointed to the angle list installed with the package in the conda environment. Installing this new version makes the pointers invalid, so perhaps this just prevents backwards compatibility. Which would be fine as its only needed for the extraction. =

src/pytom_tm/tmjob.py
sroet: I decided to regenerate it on the fly instead of storing it to json and also restructured the way the loading logging is done

sroet:
@McHaillet  thank you for your review. I restructured a bit of the code with the logging.
>  Now some particles got higher scores and others slightly lower. 

I wonder if that is because we are not generating the "poles" anymore? should we add those explicitly?

McHaillet:
> I wonder if that is because we are not generating the "poles" anymore? should we add those explicitly?

@sroet I was first thinking it would not make a big difference, but then realised it would be two sphere points but with all the in-plane rotations, correct? 

And the points closest to the poles are always the most 'unevenly' distributed?

sroet:
> @sroet I was first thinking it would not make a big difference, but then realised it would be two sphere points but with all the in-plane rotations, correct?

Yeah, that's indeed correct

> And the points closest to the poles are always the most 'unevenly' distributed?

Yeah, those are the ones that are hard to approximate with a band of squares

McHaillet (approved):
LGTM!

I would be okay to merge, we could still add the 'pole' angles in a later PR and make it an issue.

I could also still test it on my 'benchmark' project before merging. But that will be in the afternoon.

sroet:
> I would be okay to merge, we could still add the 'pole' angles in a later PR and make it an issue.

Adding the poles is trivial, so added it

> I could also still test it on my 'benchmark' project before merging. But that will be in the afternoon.

Please do! Interested to see if adding the poles helps (I would expect it to be significant if the template is already aligned with the most common orientation)

McHaillet:
> Please do! Interested to see if adding the poles helps (I would expect it to be significant if the template is already aligned with the most common orientation)

Will do 👍

McHaillet:
Got some race condition again on my home laptop:
```
........
======================================================================
FAIL: test_tm_job_split_volume (test_tmjob.TestTMJob.test_tm_job_split_volume)
----------------------------------------------------------------------
  File "/home/marten/coding/pytom-template-matching-gpu/tests/test_tmjob.py", line 301, in test_tm_job_split_volume
    self.assertAlmostEqual(angle_diff, 0, places=1, msg='angle diff should not change')
AssertionError: 103.0 != 0 within 1 places (103.0 difference) : angle diff should not change

----------------------------------------------------------------------
Ran 35 tests in 16.380s

FAILED (failures=1)
```

Could it have something to do with the poles and/or sorting?

McHaillet:
Poles are not making a difference:

angular_search_comparison_poles

src/pytom_tm/angles.py
```diff
@@ -1,26 +1,61 @@
...
+    used_npix = hp.nside2npix(nside)
+    used_angle_diff = (4*np.pi/used_npix)**0.5 * (180/np.pi)
+    logging.log(log_level, f"Using an angle difference of {used_angle_diff:.4f} for Z1 and X")
+    theta, phi = hp.pix2ang(nside, np.arange(used_npix))
+    # Explicitly add the poles; healpix does not generate these 
+    # They might be significant if template is the most occuring orientation
+    theta = itt.chain([0.0, np.pi], theta)
+    phi = itt.chain([0.0, 0.0], phi)
+    # Now for psi
+    n_psi_angles = int(np.ceil(360/angle_diff))
```
McHaillet, line 52:
```suggestion
    n_psi_angles = int(np.ceil(360/used_angle_diff))
```

Should we change this to used_angle_diff, so that the psi angles match the healpix sampling as closely as possible?

sroet:
> Poles are not making a difference:

Let's strip them out then, as they might have introduces a race condition (depending of the symmetry of the test)

src/pytom_tm/angles.py
sroet: I wanted to get as close to the input angle diff as possible (reduce search space) while still being under. 
So I have a small preference for the current one (we can't make them equal anyway)

sroet:
@McHaillet I reverted the addition of the poles and responded to your suggestion, please have another look

McHaillet:
> Got some race condition again on my home laptop:
> 
> ```
> ........
> ======================================================================
> FAIL: test_tm_job_split_volume (test_tmjob.TestTMJob.test_tm_job_split_volume)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/marten/coding/pytom-template-matching-gpu/tests/test_tmjob.py", line 301, in test_tm_job_split_volume
>     self.assertAlmostEqual(angle_diff, 0, places=1, msg='angle diff should not change')
> AssertionError: 103.0 != 0 within 1 places (103.0 difference) : angle diff should not change
> 
> ----------------------------------------------------------------------
> Ran 35 tests in 16.380s
> 
> FAILED (failures=1)
> ```
> 
> Could it have something to do with the poles and/or sorting?

I add some prints to figure out where it fails. It's only one rotation in the split volume test that is different from the reference:

```
locations with diffs: (30, 14, 26)
index in split: 447
angle in split: (5.497787143782138, 0.4111378623223477, 4.39822971502571)
index in ref: 344
angle in ref: (3.9269908169872414, 1.9106332362490186, 2.5132741228718345)
```

Position (30, 14, 26) is not close to the template position, which should be (77, 26, 40). 
So unclear what goes wrong.

McHaillet:
> @McHaillet I reverted the addition of the poles and responded to your suggestion, please have another look

Error in the test persists :/

sroet:
> > @McHaillet I reverted the addition of the poles and responded to your suggestion, please have another look
> 
> Error in the test persists :/

Wait, it wasn't just introduced with the poles? You are sure you are running a clean install of this branch?

McHaillet:
I will double check.

McHaillet:
Complete fresh install, still happens. No clue. When it tested the branch before holiday here, the tests worked fine. I could try to go back to that commit.

sroet:
> Complete fresh install, still happens. No clue. When it tested the branch before holiday here, the tests worked fine. I could try to go back to that commit.

Could you also try commit [79efbec](https://github.com/SBC-Utrecht/pytom-match-pick/pull/163/commits/79efbecdc80c4b5496152d72f88ee9d5762c65f5)  and the current main branch?

McHaillet:
> > Complete fresh install, still happens. No clue. When it tested the branch before holiday here, the tests worked fine. I could try to go back to that commit.
> 
> Could you also try commit [79efbec](https://github.com/SBC-Utrecht/pytom-match-pick/pull/163/commits/79efbecdc80c4b5496152d72f88ee9d5762c65f5) and the current main branch?

Exactly, just tried [79efbec](https://github.com/SBC-Utrecht/pytom-match-pick/pull/163/commits/79efbecdc80c4b5496152d72f88ee9d5762c65f5) and it also fails. I am 100% sure it worked before.

Will also try main.

McHaillet:
> Will also try main.

Argg, there the tests work

sroet:
Alright, so something got triggered in with the changing of the angles.
I am very surprised it is in the volume split test, not the angle split test.

Can you see if those two rotations actually have the same CC score after the volume split, or if we are now somehow catching an artifact of the volume splitting?

sroet:
Also, are you using the `conda-forge` or the `pip` version of cupy? If you are using the `pip` version, does it change if you switch to the one on `conda-forge`?

sroet:
I also get an error on the same test, but with a different angle
```python
======================================================================
FAIL: test_tm_job_split_volume (test_tmjob.TestTMJob.test_tm_job_split_volume)
----------------------------------------------------------------------
  File "/home/sander/github_files/pytom-template-matching-gpu/tests/test_tmjob.py", line 301, in test_tm_job_split_volume
    self.assertAlmostEqual(angle_diff, 0, places=1, msg='angle diff should not change')
AssertionError: 192.0 != 0 within 1 places (192.0 difference) : angle diff should not change

----------------------------------------------------------------------
Ran 35 tests in 18.498s

FAILED (failures=1)
```

McHaillet:
> I also get an error on the same test, but with a different angle
> 
> ```python
> ======================================================================
> FAIL: test_tm_job_split_volume (test_tmjob.TestTMJob.test_tm_job_split_volume)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/sander/github_files/pytom-template-matching-gpu/tests/test_tmjob.py", line 301, in test_tm_job_split_volume
>     self.assertAlmostEqual(angle_diff, 0, places=1, msg='angle diff should not change')
> AssertionError: 192.0 != 0 within 1 places (192.0 difference) : angle diff should not change
> 
> ----------------------------------------------------------------------
> Ran 35 tests in 18.498s
> 
> FAILED (failures=1)
> ```

Is it the same location?

McHaillet:
> Also, are you using the `conda-forge` or the `pip` version of cupy? If you are using the `pip` version, does it change if you switch to the one on `conda-forge`?

Am using the pip version, will give conda forge a try as well

sroet:
> Is it the same location?

Using the following code:
```python
test_diff = np.abs(
    angle[ok_region, ok_region, ok_region] -
    read_mrc(TEST_ANGLES)[ok_region, ok_region, ok_region])
print(np.argwhere(test_diff > 0.5))
```

I get
```python
[[40 20  9]]
```

McHaillet:
So also a single location, but different position

McHaillet:
Installed the conda forge cupy (its the same version as via pip, i.e. 13.1).

I get the same error and same location as I had before (but different from yours clearly):

[[30 14 26]]

For full clarity, these are my installed package versions:

```
Package             Version
------------------- -----------
contourpy           1.2.1
cupy                13.1.0
cycler              0.12.1
fastrlock           0.8.2
fonttools           4.51.0
healpix             2024.1
importlib_resources 6.4.0
kiwisolver          1.4.5
matplotlib          3.9.0
mrcfile             1.5.0
numpy               1.26.4
packaging           24.0
pandas              2.2.2
pillow              10.3.0
pip                 24.0
pyarrow             16.1.0
pyparsing           3.1.2
python-dateutil     2.9.0.post0
pytom-match-pick    0.7.0
pytz                2024.1
scipy               1.13.0
seaborn             0.13.2
setuptools          70.0.0
six                 1.16.0
starfile            0.5.8
tqdm                4.66.4
typing_extensions   4.11.0
tzdata              2024.1
voltools            0.6.0
wheel               0.43.0
```

sroet:
So I found a single split in which it does not happen, could you check if this is also a working split for you? (might tell us what is wonky in the algorithm)
```
Failing    Passing
(2,3,2)    (3,3,2)
(1,3,2)    
(2,1,2)
(2,3,1)
(2,2,2)
(2,4,2)
(2,3,3)

**Comment by McHaillet** [comment]
For me its this:

```
Failing: (2,3,2)
Passing: (1,3,2),(2,1,2),(2,3,1),(2,2,2),(2,4,2),(2,3,3),(3,3,2)
```

sroet:
It becomes weirder...
Doing a complete clean conda install (mimicking our test installation), it now does not fail for `(2,3,2)` but it does fail for `(3,3,2)`

sroet:
Alright, it the diff seems to be in generated in the actual running of the template matching.

I made sure that every output in the angles and scores matrices is written to **once** so it is not a race condition in the merging of the data

McHaillet:
> Alright, it the diff seems to be in generated in the actual running of the template matching.
> 
> I made sure that every output in the angles and scores matrices is written to **once** so it is not a race condition in the merging of the data

Good that the merging is not causing problems.

To sum up:
- It happens in a (single) specific location that happens only for some splits, but the location does not change with splits
- The location is not close to any split boundary
- It varies per machine (?)

My hunch is: at this location the template has two orientations that correlate almost identically. As the splitting changes the volumes very slightly, it occurs sometimes that it changes the correlation there slightly. Could it then just be dependent on the random number generation?

sroet:
@McHaillet with the refactored volume spitting code in #170  I can't get my test environment to error out for `(3,3,2)` anymore. Please have another run of it :crossed_fingers: :

McHaillet:
That sounds promising!! Will test!

sroet:
> To sum up, the FFT shape aware splitting is in general a good addition, and at this point we did a lot to fix this tiny inconsistency. I would we say we relax the test for volume splitting to tolerate a small difference.

I commented out the angle diff test (and added a comment why), is that a suitable solution for you?

src/pytom_tm/angles.py
McHaillet: That makes sense 👍

McHaillet (approved):
> I commented out the angle diff test (and added a comment why), is that a suitable solution for you?

That is fine for me!

LGTM! Feel free to merge :)

PR #165, opened 2024-05-03
small speedboost by precalculating conjugation

### LINKED ISSUE #103: unnecessay recalculation of rfft of mask in mean_under_mask_convolution
The rfftn() and conj() can be precalculated in std_under_mask_convolution() to save an Fourier transform (https://github.com/SBC-Utrecht/pytom-template-matching-gpu/blob/87970b30ae29f3245278ae37be3b2a477b5d6204/src/pytom_tm/matching.py#L219)

Only relevant for non-spherical mask option as there every iterations requires recalculation of std_v.

McHaillet:
PR to improve the speed a bit more and setup a few things to allow easier implementation of phase randomization.

Conjugated rft's of the tomogram are precalculated (instead of template conjugation). This requires some different shifting of the output score map and a different mask for calculating the score map standard deviation (hence some changes there).

Results are consistent with previous implementation and all the tests pass.

also closes #103

### DIFF
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -166,14 +166,14 @@ class TemplateMatchingGPU:
         cxv, cyv, czv = sxv // 2, syv // 2, szv // 2
 
         # create slice for padding
-        shift = cp.floor(cp.array(self.plan.scores.shape) / 2).astype(int) + 1
         pad_index = (
             slice(cxv - cxt, cxv + cxt + mx),
             slice(cyv - cyt, cyv + cyt + my),
             slice(czv - czt, czv + czt + mz),
         )
 
-        # calculate roi size
+        # calculate roi mask
+        shift = cp.floor(cp.array(self.plan.scores.shape) / 2).astype(int) + 1
         roi_mask = cp.zeros(self.plan.scores.shape, dtype=bool)
         roi_mask[self.stats_roi] = True
         roi_mask = cp.flip(cp.roll(roi_mask, -shift, (0, 1, 2)))
@@ -250,11 +250,7 @@ class TemplateMatchingGPU:
                 self.plan.angles
             )
 
-            self.stats['variance'] += (
-                masked_square_sum(
-                    self.plan.ccc_map, roi_mask
-                ) / roi_size
-            )
+            self.stats['variance'] += masked_square_sum(self.plan.ccc_map, roi_mask, roi_size)
 
         # get correct orientation back
         self.plan.scores = cp.roll(cp.flip(self.plan.scores), shift, axis=(0, 1, 2))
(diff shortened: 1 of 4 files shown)


src/pytom_tm/matching.py
```diff
@@ -189,16 +208,14 @@ def run(self) -> tuple[npt.NDArray[float], npt.NDArray[float], dict]:
                     output=self.plan.mask,
                     rotation_units='rad'
                 )
-                self.plan.mask_padded[cxv - cxt:cxv + cxt + mx,
-                                      cyv - cyt:cyv + cyt + my,
-                                      czv - czt:czv + czt + mz] = self.plan.mask
+                self.plan.mask_padded[pad_index] = self.plan.mask
```
sroet, line 211: these lines https://github.com/SBC-Utrecht/pytom-match-pick/blob/1fbbfb6ea8ecbe969ac558fa189e61118890e742/src/pytom_tm/matching.py?plain=1#L205-L218 are not covered by tests, please add a test (even if it wasn't before)

src/pytom_tm/matching.py
```diff
@@ -36,42 +36,45 @@ def __init__(
             reduced form, with dimensions (sx, sx, sx // 2 + 1)
         """
         # Search volume + and fft transform plan for the volume
-        self.volume = cp.asarray(volume, dtype=cp.float32, order='C')
-        self.volume_rft = rfftn(self.volume)
+        volume_shape = volume.shape
+        self.volume_rft_conj = rfftn(cp.asarray(volume, dtype=cp.float32, order='C')).conj()
+        self.volume_sq_rft_conj = rfftn(cp.asarray(volume, dtype=cp.float32, order='C') ** 2).conj()
```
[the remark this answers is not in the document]
McHaillet: Thanks for the spot!
The speed boost is mainly influenced by the angular search loop, which is where the preconjugation helps :)

src/pytom_tm/matching.py
```diff
@@ -262,66 +284,40 @@ def run(self) -> tuple[npt.NDArray[float], npt.NDArray[float], dict]:
 
 
 def std_under_mask_convolution(
-        volume: cpt.NDArray[float],
+        volume_rft_conj: cpt.NDArray[float],
+        volume_sq_rft_conj: cpt.NDArray[float],
```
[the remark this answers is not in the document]
McHaillet: will do

src/pytom_tm/matching.py
McHaillet: that is for non-spherical masks, i'll make sure they are covered, should be a very quick fix

sroet: this is still not hit

sroet (changes requested):
still 1 missing test, LGTM otherwise

McHaillet:
Woops this didn't fix it yet

tests/test_template_matching.py
```diff
@@ -31,8 +31,49 @@ def test_search(self):
...
+
+        tm = TemplateMatchingGPU(
+            0,
+            0,
+            self.volume,
+            self.template,
+            self.mask,
+            self.angles,
+            list(range(len(self.angles))),
+            mask_is_spherical=True,
```
McHaillet, line 75:
```suggestion
            mask_is_spherical=False,
```

src/pytom_tm/matching.py
McHaillet: Set the wrong value haha
Hope it is fixed now

McHaillet:
I set precision to 4 for the standard deviation. The although the spherical the interpolation on rotations still slightly changes the mask, meaning the eventual scores won't be exactly identical. It should still be very close and that is indeed true.

McHaillet:
Its fixed now. The lines are hit!

McHaillet:
Needed to merge changes from main to make it mergeable. Lets see if they now pass.

PR #168, opened 2024-05-07
wait longer to prevent leaked processes

McHaillet:
Changed from 1 to 2 seconds. You think that should be sufficient?


sroet (approved):
test pass, so seems sufficient for now :sweat_smile: , LGTM

McHaillet:
Let's see if it holds

PR #169, opened 2024-05-08
make template non-symmetrical for template matching test to prevent race condition

McHaillet:
I think this should solve it.

I'll wait for the tests to pass


sroet (approved):
LGTM! thanks for looking into this! feel free to merge on green

McHaillet:
standard deviation changed, so had to update the test values

PR #170, opened 2024-05-24
refactor volume splitting

sroet:
This makes the volume splitting FFT aware and tries to be somewhat smart about it (including more of the original data instead of padding).

Hopefully preventing the testing artifacts seen in #163 

TODO:

- [X] refactor volume splitting 
- [ ] update docstrings
- [ ] add type hints to added functions
- [ ] make sure everything is tested

### DIFF
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -3,6 +3,7 @@ from importlib import metadata
 from packaging import version
 import pathlib
 import copy
+import itertools as itt
 import numpy as np
 import numpy.typing as npt
 import json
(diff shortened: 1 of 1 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -79,6 +80,92 @@ def load_json_to_tmjob(file_name: pathlib.Path, load_for_extraction: bool = True
...
+            elif no_overhang_left + min_len >= length:
+                # Last slice, only overhang to the left
+                split_length = next_fast_len(min_len)
+                data_slices.append((length-split_length, length))
+                valid_data_slices.append((length-split_length+overhang, length))
+                sub_len.append(split_length)
+                break
+            else:
+                # Any other slice
+                split_length = next_fast_len(min_len+overhang)
```
McHaillet, line 133: I don't think I fully understand this part. Does `min_len` not already include the overhang? Why is it again added for the central slices?

McHaillet (changes requested):
Looks nice, and it simplifies the job splitting code a lot :)

Just one question about something that is not fully clear to me!

src/pytom_tm/tmjob.py
sroet: It includes a single overhang (left **or** right), but we want it twice for any center slice (left **and** right)

McHaillet: ah! 

I got confused because before I had the overhang as template_size//2. But for the function that does not matter of course! The overhang can be set to whatever.

In that case, could you add back in on lines 467-469, that the input overhang is (template_size//2) + 1 (for uneven shape) ? Due to the overlap from both sides the full region will already be covered.

McHaillet (approved):
LGTM! :+1:

src/pytom_tm/tmjob.py
```diff
@@ -79,6 +81,101 @@ def load_json_to_tmjob(file_name: pathlib.Path, load_for_extraction: bool = True
...
+            unique_data.append((unique_left, length))
+        # Make sure unique slices are unique and within valid data
+        last_right = 0
+        for (vd_left, vd_right),(ud_left, ud_right) in zip(valid_data_slices, unique_data):
+            if (ud_left < vd_left or 
+                ud_right > vd_right or
+                ud_right > length or
+                ud_left != last_right
+                ): # pragma: no cover
+                raise RuntimeError(f"We produced inconsistent slices for {length=}, {splits=}, {overhang=}")
```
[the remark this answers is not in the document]
sroet: I am pretty sure a too large overhang results in a negative `no_overhang_left` which trips the error check on line `149`

Issue #173, opened 2024-06-03
center of mass not working as expected


McHaillet:
Maybe this centering option should be reevaluated. I had some small issues with it before. This is the scipy docs for center of mass (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.center_of_mass.html) the bottom lines say it can return a RuntimeWarning and return inf or nan values, which might be why your template ends up empty.

_Originally posted by @McHaillet in https://github.com/SBC-Utrecht/pytom-match-pick/discussions/172#discussioncomment-9647246_

sroet:
From an offline discussion: Looking at a center of mass code, it looks like it is not behaving as we expect when both positive and and negative voxels are present. We decided to use array**2 to compute the center of mass instead (hopefully also focusing more on the strong data vs noise)

PR #174, opened 2024-06-04
Ctf rework

### LINKED ISSUE #175: write PSF and convolved template in debug mode of match_template

McHaillet:
to-do:
- [x] phase flip and check contrast
- [x] remove ctf parameter input from create_template entry point
- [x] add test for template creation

create_template.py no longer supports input parameters for CTFs with this update. CTF's can only be applied in match_template.py. I think this simplifies the user input.

To make this easier, I changed the CTF output so that the user only needs to ensure the contrast of the reference matches the contrast of the tomogram (i.e. black tomogram means black template, and vice versa). After CTF convolution the template should then always get the right contrast. I also made issue #175 because that should still gives people an option to inspect if the weighting is internally correct.

I did not update version number as we did not tag an official release for 0.7.0 yet.

closes #175


McHaillet:
I should also update the docs accordingly with this, and for the angular sampling update as well.

src/pytom_tm/io.py
```diff
@@ -100,18 +100,33 @@ def __call__(self, parser, namespace, values: str, option_string: Optional[str]
...
+            parser,
+            namespace,
+            values: str,
+            option_string: Optional[str] = None
+    ):
+        if values.endswith(('.defocus', '.txt')):
+            file_path = pathlib.Path(values)
+            if not file_path.exists():
+                parser.error(
+                    "{0} provided defocus file does not exist".format(option_string)
```
sroet, line 117: I prefer f-strings
```suggestion
                    f"{option_string} provided defocus file does not exist"
```

src/pytom_tm/io.py
```diff
@@ -100,18 +100,33 @@ def __call__(self, parser, namespace, values: str, option_string: Optional[str]
...
+                )
+            # in case of a file the parser attribute becomes a list of defocii
+            setattr(namespace, self.dest, read_defocus_file(file_path))
+        else:
+            try:
+                defocus = float(values)
+            except ValueError:
+                parser.error(
+                    "{0} not possible to read defocus as float".format(option_string)
```
sroet, line 126: I prefer f-strings
```suggestion
                    f"{option_string} not possible to read defocus as float"
```

src/pytom_tm/entry_points.py
```diff
@@ -831,7 +758,7 @@ def match_template(argv=None):
 
     # combine ctf values to ctf_params list of dicts
     ctf_params = None
-    if args.defocus_file is not None:
+    if args.defocus is not None:
```
sroet, line 761: This whole function is untested, would it make sense to add tests to this entry point in this PR as well?

src/pytom_tm/io.py
```diff
@@ -100,18 +100,33 @@ def __call__(self, parser, namespace, values: str, option_string: Optional[str]
...
-    def __call__(self, parser, namespace, values: str, option_string: Optional[str] = None):
-        file_path = pathlib.Path(values)
-        if not file_path.exists():
-            parser.error("{0} provided defocus file does not exist".format(option_string))
-        allowed_suffixes = ['.defocus', '.txt']
-        if file_path.suffix not in allowed_suffixes:
-            parser.error("{0} provided defocus file does not have the right suffix, "
-                         "allowed are: {1}".format(option_string, ', '.join(allowed_suffixes)))
-        setattr(namespace, self.dest, read_defocus_file(file_path))
+class ParseDefocus(argparse.Action):
```
sroet, line 103: this added function is untested, please add a test

McHaillet:
Decided I would immediately add issue #175

closes #175

sroet:
> Decided I would immediately add issue #175
> 
> closes #175

You probably have to add `closes #175` to the opening comment

tests/test_template.py
```diff
@@ -0,0 +1,52 @@
...
+            self.template, 1, 1, center=False,
+        )
+        square_sum = np.square(new_template - self.template).sum()
+        self.assertTrue(square_sum < 1, msg="Template should not change strongly "
+                                            "without recentering.")
+        new_template = generate_template_from_map(
+            self.template, 1, 1, center=True,
+        )
+        square_sum = np.square(new_template - self.template).sum()
+        self.assertTrue(square_sum > 1, msg="Template didnt change after shift")
```
sroet, line 38: Do you think it makes sense to make sure this test tests mixed high and low values in the template? (to catch #173)

sroet (changes requested):
Looks reasonable, had a couple comments; feel free to argue against them

Other things:
- testing should be upped to 82

tests/test_template.py
McHaillet: My idea was to add those things in for that PR and here just prep the test for that one.

src/pytom_tm/entry_points.py
McHaillet: yea, could do it. Do you have any examples of proper entry point testing with all arguments and stuff?

tests/test_template.py
sroet: :+1:

src/pytom_tm/entry_points.py
sroet: It is quite trivial, you just pass it a list of arguments that are split on spaces
so `['-t', template_path, '-v', tomogram_path]` etc or you can just write the string with command options (not the entry point name itself) and use `commands.split()`

McHaillet: This is added

src/pytom_tm/io.py
McHaillet: The input is tested via the entry point. I don't want to spend time right investigating how to write a test around these custom actions.

src/pytom_tm/tmjob.py
```diff
@@ -684,6 +684,18 @@ def start_job(
...
+                write_mrc(
+                    self.output_dir.joinpath("template_psf.mrc"),
+                    template_wedge,
+                    self.voxel_size,
+                )
+                write_mrc(
+                    self.output_dir.joinpath("template_convolved.mrc"),
+                    irfftn(rfftn(template) * template_wedge, s=template.shape),
+                    self.voxel_size,
+                )
```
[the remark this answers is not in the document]
McHaillet: still needs to be done

McHaillet:
Pushed some updates, coverage is much better. However terminal is flooded quite a lot during the entry point test. Maybe you have an idea how to solve it?

src/pytom_tm/io.py
sroet: Hitting it with the entry point is perfectly fine!

sroet:
> Maybe you have an idea how to solve it?

I will have a look!

McHaillet:
I think something goes wrong with the logging level. By changing it in the entry point, it might also affect the unittest logging.

tests/test_entry_points.py
```diff
@@ -1,8 +1,11 @@
 import unittest
+import pathlib
+import numpy as np
 from shutil import which
 from contextlib import redirect_stdout
```
sroet, line 5:
```suggestion
from contextlib import redirect_stdout, redirect_stderr
```

tests/test_entry_points.py
```diff
@@ -34,3 +79,54 @@ def test_entry_points_exist(self):
...
+        # test valid defocus arguments
+        for z in [str(DEFOCUS_IMOD), str(DEFOCUS), '3000']:
+            arguments = defaults.copy()
+            arguments['--defocus'] = z
+            start(arguments)
+
+        # test faulty args
+        for z in ['asdf.txt', 'asdf']:
+            dump = StringIO()
+            with self.assertRaises(SystemExit) as ex, redirect_stdout(dump):
```
sroet, line 114: you also have to dump stderr (with the import above
```suggestion
            with self.assertRaises(SystemExit) as ex, redirect_stdout(dump), redirect_stderr(dump):
```

src/pytom_tm/tmjob.py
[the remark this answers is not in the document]
McHaillet: Actually, not working when I run the entry point with debug mode. I have pushed my last commit for this, but its not working. Maybe related to the thing below?

sroet (commented):
this fixes the clutter from the entry points on my test system

src/pytom_tm/tmjob.py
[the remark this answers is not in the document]
sroet: From: https://docs.python.org/3/library/logging.html#logging.basicConfig
> This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True.

So we should probably set `force=True` in where we set up the logger (I think resetting any log handlers should be intended behavior of entry points)
If this is not intended then we should probably do
```python
if logging.getLogger().handlers:
    # basicConfig has been called before
    logging.getLogger().setLevel(args.log)
else:
    # set up base logger
    logging.basicConfig(level=args.log)
```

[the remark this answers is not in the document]
McHaillet: > From: https://docs.python.org/3/library/logging.html#logging.basicConfig
> 
> > This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True.
> 
> So we should probably set `force=True` in where we set up the logger (I think resetting any log handlers should be intended behavior of entry points) If this is not intended then we should probably do

That seems fair for the entry points

McHaillet:
Problem seems to be that if I use `force` the logging is changed for all the unittests that come after.

sroet:
> Problem seems to be that if I use force the logging is changed for all the unittests that come after.

Yes, that is kinda intended with force indeed.
However, shouldn't all tests that depend on logging anyway have a filter set up that filter to the correct level? (or explicitly call `basicConfig(level=logging.INFO, force=True)` itself? )

McHaillet:
Once the logging is set to debug, all tmjobs called after will also write the debug files that I added. And that messes with expected output. 
Perhaps it's just not a good idea to write files dependent on the logging level.

McHaillet:
I though maybe with a context manager for the logging level, but I don't think it exists

McHaillet:
I now just set it to debug for this specific test. Let me know if you agree!

sroet (changes requested):
Thanks for all the fixes

One final comment about the changed debug test, LGTM otherwise.

tests/test_entry_points.py
```diff
@@ -34,3 +80,63 @@ def test_entry_points_exist(self):
...
+        with self.assertRaises(ValueError, msg='Missing CTF params should produce '
+                                               'error'):
+            arguments = defaults.copy()
+            arguments.pop('--voltage')
+            start(arguments)
+
+        # test debug files
+        arguments = defaults.copy()
+        start_level = logging.getLogger().level
+        logging.basicConfig(level=10, force=True)  # set to debug
```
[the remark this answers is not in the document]
McHaillet: agreed!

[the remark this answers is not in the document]
McHaillet: its solved

sroet (approved):
LGTM, feel free to merge

McHaillet:
Thanks!

PR #178, opened 2024-06-05
Relion5 compatibility

### LINKED ISSUE #177: extract should support relion5 particle annotations
This means each positions should be subtracted by tomo.shape // 2 as relion5 defines particle positions relative to the center of the tomogram. extract_candidates should get an additional command line argument for output coordinates in this fashion.

McHaillet:
closes #177 

Coordinates are written relative to the tomogram center and in Angstrom (instead of nr of voxels). Some star file column header names also changed. 

I checked with relion what definition if of the tomogram center: 3dem/relion#1137

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -465,11 +465,11 @@ def extract_candidates(argv=None):
              "leave at 1."
     )
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


sroet (approved):
LGTM! feel free to merge!

PR #179, opened 2024-06-05
Center of mass fix

### LINKED ISSUE #173: center of mass not working as expected

McHaillet:
Fix for the center of mass. I am not fully sure about the unittest check yet though, maybe you have a better solution. Currently checking with the same method as I am using internally.

I also switched to box padding after recentering. That is the better order as the padding is anyway done with zeros.

Closes #173

The square seems to work nicely though. This is an example that was failing before:
 
centering_before_fix

And not works much better:

centering_after_fix

### DIFF
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -75,8 +75,8 @@ def generate_template_from_map(
         logging.debug(
             f'size check {output_box_size} > {(input_map.shape[0] * input_spacing) // output_spacing}')
         if output_box_size > (input_map.shape[0] * input_spacing) // output_spacing:
-            pad = int(output_box_size * (output_spacing / input_spacing)) - \
-                  input_map.shape[0]
+            pad = (int(output_box_size * (output_spacing / input_spacing)) - 
+                   input_map.shape[0])
             logging.debug(f'pad with this number of zeros: {pad}')
             input_map = np.pad(
                 input_map,
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -59,11 +59,24 @@ def generate_template_from_map(
         logging.warning(warning_text)
         filter_to_resolution = 2 * output_spacing
 
+    if center:
+        volume_center = np.divide(np.subtract(input_map.shape, 1), 2, dtype=np.float32)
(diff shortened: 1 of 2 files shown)


src/pytom_tm/template.py
```diff
@@ -59,32 +59,37 @@ def generate_template_from_map(
...
+
     # extend volume to the desired output size before applying convolutions!
     if output_box_size is not None:
-        logging.debug(f'size check {output_box_size} > {(input_map.shape[0] * input_spacing) // output_spacing}')
+        logging.debug(
+            f'size check {output_box_size} > {(input_map.shape[0] * input_spacing) // output_spacing}')
         if output_box_size > (input_map.shape[0] * input_spacing) // output_spacing:
-            pad = int(output_box_size * (output_spacing / input_spacing)) - input_map.shape[0]
+            pad = int(output_box_size * (output_spacing / input_spacing)) - \
+                  input_map.shape[0]
```
sroet, line 79: I prefer parenthesis over backslashes
```suggestion
            pad = (int(output_box_size * (output_spacing / input_spacing)) - 
                   input_map.shape[0])
```

sroet (approved):
One preference, feel free to ignore. LGTM otherwise

>  I am not fully sure about the unittest check yet though, maybe you have a better solution.

Current implementation looks good
One other solution might be to compare `np.abs(generate_template_from_map(template, ...)` and `generate_template_from_map(np.abs(template), ...)` which should be identical

McHaillet:
> Current implementation looks good One other solution might be to compare `np.abs(generate_template_from_map(template, ...)` and `generate_template_from_map(np.abs(template), ...)` which should be identical

I added this in. Will merge once the tests pass.

PR #181, opened 2024-06-06
calculate angular search from particle diameter

### LINKED ISSUE #164: provide a feature to go from the particle diameter to the 'correct' sampling and the pixel size (or low pass filter)
>Very excited to see this one!
We could also provide a feature to let users input the particle diameter, and calculatie the 'correct' sampling from that and  the pixel size (or low pass filter)

_Originally posted by @McHaillet in https://github.com/SBC-Utrecht/pytom-match-pick/issues/163#issuecomment-2093078631_

McHaillet:
Closes #164 

Addition of particle diameter to calculate angular search.

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -548,11 +548,24 @@ def match_template(argv=None):
     )
     rotation_group = parser.add_argument_group('Angular search')
     rotation_group.add_argument(
+        "--particle-diameter",
+        type=float,
+        required=True,
+        action=LargerThanZero,
+        help="Provide a particle diameter to automatically determine the ideal "
+             "angular sampling using the Crowther criterion. For the max resolution "
+             "the pixel size is used unless a low-pass filter is specified, in which "
+             "the low-pass resolution is used. For non-globular macromolecules choose "
+             "the diameter along the longest axis.",
+    )
+    rotation_group.add_argument(
         "--angular-search",
         type=str,
         required=True,
-        help="If given a float it will generate an angle list with healpix for Z1 and X1 "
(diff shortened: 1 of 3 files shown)


src/pytom_tm/entry_points.py
```diff
@@ -555,12 +555,25 @@ def match_template(argv=None):
...
     )
     rotation_group = parser.add_argument_group('Angular search')
+    rotation_group.add_argument(
+        "--particle-diameter",
+        type=float,
+        required=False,
+        action=LargerThanZero,
+        help="Provide a particle diameter (in Angstrom) to automatically determine the "
+             "angular sampling using the Crowther criterion. For the max resolution "
+             "the pixel size is used unless a low-pass filter is specified, in which "
```
sroet, line 565: it is pixelsize * 2, right?
```suggestion
             "the 2 * pixel size is used unless a low-pass filter is specified, in which "
```

src/pytom_tm/tmjob.py
```diff
@@ -332,6 +336,13 @@ def __init__(
         self.start_slice = 0
         self.steps_slice = 1
         self.rotational_symmetry = rotational_symmetry
+        self.particle_diameter = particle_diameter
+        # calculate increment from particle diameter
+        if angle_increment is None and particle_diameter is not None:
```
sroet, line 341: should we update the default angle increment to `None` instead of `7.0` and set it to `7.0` during the init?

sroet (approved):
one help question and a suggestion for updated defaults, feel free to ignore. LGTM otherwise

McHaillet:
I'm merging!

Issue #171, opened 2024-05-25
Numpy 2.0 fft updates might be relevant


McHaillet:
Just a reminder to check if it's relevant:
https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-fft-deprecations-for-n-d-transforms-with-none-values-in-arguments

McHaillet:
This should be updated. Essentially all the irfftn's I used only have the `s` argument and not the `axes` argument. Can I guess be fixed like so:

```
ft = np.fft.rfftn(image)
new_image = np.fft.irfftn(ft, s=image.shape, axes=list(range(image.shape)))
```

sroet:
So from a glance none of our files are affected (so far), I `grep`ed all python files containing :
```
matching.py #only cupy based
weights.py #only np.fft.fftshift and np.fft.ifftshift, not deprecated
template.py # only scipy based
tmjob.py # only scipy based
```

So I think this issue can be closed?

McHaillet:
you're right! Is there a chance that scipy will follow numpy2 convention?

sroet:
Maybe at some point, but I kinda expect them to just implement the `list(range(image.shape))` themselves :crossed_fingers:

McHaillet:
Okay, let's close it!

PR #182, opened 2024-06-06
phase randomization score map correction

McHaillet:
This PR adds 'phase randomization', introduced by STOPGAP.

Explanation of what this is: The match_template command line got two additional arguments: one flag to activate this mode, another to set a seed for randomization. The seed is used to create a phase randomized version of the template, meaning the phases of the Fourier transform are randomly permuted (used often for FSC calculations). This second template is like a random noise object and is matched simultaneously with the original template. The score map of the noisy version is subtracted from the real one.

With the particle diameter updated and ctf input reworked, this is the last thing I really want to add in functionality-wise.

Once this one is merged, I would like make a PR to throw the code through ruff and maybe set up a pre-commit thingy.

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -758,6 +758,7 @@ def match_template(argv=None):
     )
     additional_group = parser.add_argument_group('Additional options')
     additional_group.add_argument(
+        "-r",
         "--random-phase-correction",
         action="store_true",
         default=False,
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -20,6 +20,7 @@ from pytom_tm.io import (
     BetweenZeroAndOne,
 )
 from pytom_tm.tmjob import load_json_to_tmjob
+from os import urandom
 
 
 def _parse_argv(argv=None):
@@ -771,7 +772,7 @@ def match_template(argv=None):
     additional_group.add_argument(
         "--rng-seed",
         action=LargerThanZero,
-        default=321,
+        default=int.from_bytes(urandom(8)),
         required=False,
         help="Specify a seed for the random number generator used for phase "
              "randomization for consistent results!"
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -756,6 +756,25 @@ def match_template(argv=None):
         "apply it to the tomogram patch and template. Effectively puts more weight on "
         "high resolution features and sharpens the correlation peaks.",
     )
+    additional_group = parser.add_argument_group('Additional options')
+    additional_group.add_argument(
+        "--random-phase-correction",
+        action="store_true",
+        default=False,
+        required=False,
+        help="Run template matching simultaneously with a phase randomized version of "
+             "the template, and subtract this 'noise' map from the final score map. "
+             "For this method please see STOPGAP as a reference: "
+             "https://doi.org/10.1107/S205979832400295X ."
+    )
+    additional_group.add_argument(
+        "--rng-seed",
+        action=LargerThanZero,
+        default=321,
+        required=False,
+        help="Specify a seed for the random number generator used for phase "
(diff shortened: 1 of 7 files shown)


src/pytom_tm/entry_points.py
```diff
@@ -756,6 +756,26 @@ def match_template(argv=None):
...
+        required=False,
+        help="Run template matching simultaneously with a phase randomized version of "
+             "the template, and subtract this 'noise' map from the final score map. "
+             "For this method please see STOPGAP as a reference: "
+             "https://doi.org/10.1107/S205979832400295X ."
+    )
+    additional_group.add_argument(
+        "--rng-seed",
+        action=LargerThanZero,
+        default=321,
```
sroet, line 774: I wonder if it is best to have no default for this option (generate a random seed every time this is ran)

src/pytom_tm/matching.py
```diff
@@ -101,7 +116,8 @@ def __init__(
         - pyTME: https://github.com/KosinskiLab/pyTME
 
         The precalculation of conjugated FTs of the tomo was (AFAIK) introduced
-        by STOPGAP!
+        by STOPGAP! Also, they introduced simultaneous matching with a phase randomized
+        version of the template.
```
sroet, line 120: maybe also add doi link here?
```suggestion
        version of the template. https://doi.org/10.1107/S205979832400295X
```

src/pytom_tm/matching.py
```diff
@@ -262,6 +297,51 @@ def run(self) -> tuple[npt.NDArray[float], npt.NDArray[float], dict]:
...
+                # Paste in center
+                self.plan.template_padded[pad_index] = self.plan.template
+
+                # Fast local correlation function between volume and template, norm is the standard deviation at each
+                # point in the volume in the masked area
+                self.plan.ccc_map = (
+                        irfftn(self.plan.volume_rft_conj * rfftn(self.plan.template_padded),
+                               s=self.plan.template_padded.shape)
+                        / self.plan.std_volume
+                )
```
sroet, line 330: this part is very similar to 257-285 and should in my opinion be split out into a separate function

src/pytom_tm/template.py
```diff
@@ -103,3 +108,43 @@ def generate_template_from_map(
...
+    relevant_freqs = grid <= 1  # permute only up to Nyquist
+    noise = np.zeros_like(phase)
+    rng = np.random.default_rng(seed)
+    noise[relevant_freqs] = rng.permutation(phase[relevant_freqs])
+
+    # construct the new template
+    noise = np.reshape(noise, amplitude.shape)
+    result = np.fft.irfftn(
+        amplitude * np.exp(1j * noise), s=template.shape
+    )
```
sroet, line 149: I think this should be using the imported scipy function to make sure we don't run into issues with #171 
```suggestion
    result = irfftn(
        amplitude * np.exp(1j * noise), s=template.shape
    )
```

sroet (changes requested):
general comment:
- ~min coverage should be increased to 86~ already done in #184 

Other than that, one default question, one proposed refactor, one doc suggestion, one code consistency thing

> Once this one is merged, I would like make a PR to throw the code through ruff and maybe set up a pre-commit thingy.

Sounds good, I can also look into that if you don't have time for that (isn't necessary 'publishable' :wink:  ), just tag me into an issue

src/pytom_tm/entry_points.py
McHaillet: probably, can add that

src/pytom_tm/matching.py
McHaillet: Did it!

McHaillet:
> Sounds good, I can also look into that if you don't have time for that (isn't necessary 'publishable' 😉 ), just tag me into an issue

Also fine, will respond in the PR I made for it!

sroet (approved):
LGTM, feel free to merge

PR #183, opened 2024-06-06
pre-commit with ruff

McHaillet:
PR for pre-commits.

to-do:
- [x] add pre-commit config
- [x] update .toml dev dependencies
- [x] run whole codebase through ruff check and ruff format
- [x] add ruff checks to test pipline

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -34,9 +34,9 @@ jobs:
       - name: Run tests
         run: |
           coverage run -m unittest discover tests/
-      - name: Install plotting code and list dependencies
+      - name: Install all code and list dependencies
         run: |
-          python -m pip install -e .[plotting]
+          python -m pip install -e .[all]
           conda list
       - name: Run complete tests and generate coverage report
         run: |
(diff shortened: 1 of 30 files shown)


sroet:
can I also propose adding a test that checks ruff before running the tests?

McHaillet:
@sroet I am also fine if you continue work on this PR (if you have time of course). I don't know how easy it is to take over a PR, or maybe its easier to merge this first and make a new PR?

sroet:
>  I don't know how easy it is to take over a PR, or maybe its easier to merge this first and make a new PR?

Looks like I can just push to this branch 😄 

In the last commit I added a catch-all `[all]` so that we don't have to keep updating our docs if we decide to add more optional dependencies down the line

McHaillet:
> Looks like I can just push to this branch 😄

Noice 👍 
 
> In the last commit I added a catch-all `[all]` so that we don't have to keep updating our docs if we decide to add more optional dependencies down the line

Good one

sroet:
@McHaillet I think this is ready 😄 please have a look

McHaillet:
Got it. The pipeline addition just checks if everything is formatted/linted, right?

McHaillet:
Good!

I checked the config and setup files, and looks good. I assume all the python files are auto formatted. The tests pass, so seems good from my side.

sroet (approved):

PR #184, opened 2024-06-06
fix symmetry bug

### LINKED ISSUE #180: z-axis rotational symmetry
Wouter mentioned the rotational symmetry might be incorrect. Currently it is applied along the Z1-axis, but his point was it should be applied along the Z2-axis as the ZXZ angles specify the rotation of the reference onto the particle in the tomogram. Thinking about it he is probably right.

Perhaps fix or remove argument, depending on time?

@sroet maybe you could look at it? Should I assign it?

sroet:
closes #180

### DIFF
--- a/src/pytom_tm/angles.py
+++ b/src/pytom_tm/angles.py
@@ -76,7 +76,7 @@ def load_angle_list(file_name: pathlib.Path, sort_angles: bool = True) -> list[t
         angle_list.sort()  # angle list needs to be sorted otherwise symmetry reduction cannot be used!
     return angle_list
 
-def get_angle_list(angle: Union[pathlib.Path, float], sort_angles: bool = True, log_level: str = 'DEBUG'):
+def get_angle_list(angle: Union[pathlib.Path, float], sort_angles: bool = True, symmetry: int = 1, log_level: str = 'DEBUG'):
     """Either get an angular search file from disk or generate one from a float
 
     Parameters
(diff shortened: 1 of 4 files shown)


src/pytom_tm/angles.py
```diff
@@ -88,6 +88,8 @@ def get_angle_list(angle: Union[pathlib.Path, float], sort_angles: bool = True,
           maximum difference (in degrees) for the angle list
     sort_angles: bool, default True
         sort the list, using python default angle_list.sort(), sorts first on Z1, then X, then Z2
+    symmetry: int, default 1
+        the returned list will only have Z2 angles [0, (2*pi/symmetry))
```
McHaillet, line 92: Could you add here that the symmetry is not applied for files? A similar message would also be handy to have at the command line argument for rotational symmetry.

McHaillet (changes requested):
LGTM, just one comment about docs and a suggestion for some user info.

src/pytom_tm/angles.py
sroet: But it is applied to files?
files are read in in line `118` and then filtered in line `122`?
(this was also the case in the old implementation, hence why I left it that way

McHaillet: Woops, read it wrong!

McHaillet (approved):
Good to go then

PR #185, opened 2024-06-12
Add tomogram mask option for template matching

### LINKED ISSUE #54: including mask during template matching to reduce the search space
Some people asked if the search space could be constraint during the actual template matching instead of during candidate extraction, to improve runtimes in tomograms were only a small section is of interest.

The most feasible suggestion seems to check if a patch of the tomogram (as split with the `-s` option in `pytom_match_template.py`) has values of 1 in its mask patch. If the volume is not split into patches (i.e. just one big patch), there will always be a positive mask section and therefore the search wont be restrained. However, if the volume is actually split by the user only the patches with positve mask values will have to be calculated.

sroet:
closes #54 

It will use the same definition as for the extraction `(any point <=0 is skipped)` but will initially only skip subvolumes where the complete mask is <= 0

For the extraction, it will use `job.tomogram_mask` as well, which can be overridden by giving a separate tomogram_mask in the entry point

- [x] add option to entry point
- [x] add code logic to tmjob
- [x] add extra code logic to extract
- [x] add tests

### DIFF
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -57,6 +57,8 @@ def load_json_to_tmjob(
         search_x=data["search_x"],
         search_y=data["search_y"],
         search_z=data["search_z"],
+        # Use 'get' for backwards compatibility
+        tomogram_mask=data.get("tomogram_mask", None),
         voxel_size=data["voxel_size"],
         low_pass=data["low_pass"],
         # Use 'get' for backwards compatibility
@@ -213,6 +215,7 @@ class TMJob:
         search_x: Optional[list[int, int]] = None,
         search_y: Optional[list[int, int]] = None,
         search_z: Optional[list[int, int]] = None,
+        tomogram_mask: Optional[pathlib.Path] = None,
         voxel_size: Optional[float] = None,
         low_pass: Optional[float] = None,
         high_pass: Optional[float] = None,
@@ -256,6 +259,8 @@ class TMJob:
             restrict tomogram search region along the y-axis
         search_z: Optional[list[int, int]], default None
             restrict tomogram search region along the z-axis
+        tomogram_mask: Optional[pathlib.Path], default None
+            when volume splitting tomograms, only subjobs where any(mask > 0) will be generated
         voxel_size: Optional[float], default None
             voxel size of tomogram and template (in A) if not provided will be read from template/tomogram MRCs
         low_pass: Optional[float], default None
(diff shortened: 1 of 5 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -551,6 +559,14 @@ def split_volume_search(self, split: tuple[int, int, int]) -> list[TMJob, ...]:
             )
 
         search_size = self.search_size
+        if self.tomogram_mask is not None:
+            tomogram_mask = read_mrc(self.tomogram_mask)
+            if np.all(tomogram_mask <= 0):
```
McHaillet, line 564: This should skip a full tomogram if there is no positive region in the mask. I think now its only checked split_volume_search is ran on the TMJob, and that is not always done in parallel (https://github.com/SBC-Utrecht/pytom-match-pick/blob/fcdb82c28c29fea89a2a0582edf73a03e59e3f1d/src/pytom_tm/parallel.py#L86).
Maybe better to move it to the job initialization?

tests/test_tmjob.py
```diff
@@ -86,6 +87,7 @@ def setUpClass(cls) -> None:
             overwrite=True,
         )
         write_mrc(TEST_TOMOGRAM, volume, 1.0)
+        #
```
McHaillet, line 90:
```suggestion
```

McHaillet (commented):
Nice! I only have a comment about the check for any positive value in the full mask.

src/pytom_tm/tmjob.py
sroet: I agree, decided to move it to the init of the TMJob instead

sroet:
@McHaillet did refactor to check the validity of the mask when TMJob is initialized, please have another look

src/pytom_tm/tmjob.py
```diff
@@ -363,6 +368,13 @@ def __init__(
         ]
 
         logging.debug(f"origin, size = {self.search_origin}, {self.search_size}")
+        self.tomogram_mask = tomogram_mask
+        if tomogram_mask is not None:
+            temp = read_mrc(tomogram_mask)
+            if np.all(temp <= 0):
+                raise ValueError(
+                    f"No values larger than 0 found in the template mask: {tomogram_mask}"
```
McHaillet, line 376:
```suggestion
                    f"No values larger than 0 found in the tomogram mask: {tomogram_mask}"
```

McHaillet (changes requested):
Just noticed one string that probably needs to change, LGTM otherwise :)

McHaillet (approved):
Good to go

sroet:
> Just noticed one string that probably needs to change, LGTM otherwise :)

Thanks for catching that :sweat_smile:

PR #187, opened 2024-06-14
add tutorial test yml

### LINKED ISSUE #148: Test the tutorial
We should walk through the tutorial code.
A command to pull a zip file with the 4 files for a single tomogram is:
`curl -L -O -J -H "X-Dataverse-key:$DATAVERSE_API_TOKEN" https://dataverse.nl/api/access/datafiles/384731,384724,384706,384718`

This should only run on our main branch and should probably be part of the weekly as it needs access to the repository secrets for the API token

Another option is to instead use [git lfs](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage) to store the single tomogram mrc (especially since we are now pip installable that don't package the tests anyway). This would also allow us to tests the entry points properly  and generate coverage data on the PRs for them

sroet:
closes #148 

For now this does not include coverage
It uses repo secrets for pulling from dataverse, which means it has to be spawned like the coverage comment workflow

It walks through the tutorial file, grabs all code blocks and executes them if they start with `*.py`

### DIFF
--- a/.github/workflows/tutorial-tests.yml
+++ b/.github/workflows/tutorial-tests.yml
@@ -1,0 +1,61 @@
+name: Test Tutorial
+# following example: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+on:
+  workflow_run:
+    workflows: ["Unit tests"]
+    types:
+      - completed
+
+jobs:
+  test-tutorial:
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: [self-hosted, linux, x64, gpu]
+    container:
+      image: continuumio/miniconda3
+      options: --runtime=nvidia --gpus all
+    permissions:
+      # Read the action that triggered this one
+      actions: read
+      # Allow to write check runs
+      checks: write
+      # Read and update commit statuses
+      statuses: write
+    steps:
+      - name: Pull code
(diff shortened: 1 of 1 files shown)


sroet:
@McHaillet  as with #118 this will not work properly before it is actually in the main repo due to the way the protections are set up

McHaillet:
Looks good. My main concern is that running the tutorial will take some time, a rough guess from me would be ~40 minutes (just to give a ballpark). Maybe running on every commit of a PR is too much, but perhaps it needs to pass before a version release to pypi?

McHaillet (changes requested):
Looks good. My main concern is that running all the tutorial commands takes some time, a rough guess would be ~40 mins (to have give a ballpark). Perhaps we could require the tutorial to run before releasing a new version to pypi.

sroet:
> BTW not specifically requesting to change (should have made this review comment), but just wanted to discuss this point.

Understandable, we can also just run it on the weekly builds if you want. First wanted just to see if it actually runs (and in what time)

McHaillet:
> > BTW not specifically requesting to change (should have made this review comment), but just wanted to discuss this point.
> 
> Understandable, we can also just run it on the weekly builds if you want. First wanted just to see if it actually runs (and in what time)

Alright, that sounds good :)

McHaillet (approved):

PR #197, opened 2024-07-24
fix defocus help unit

sroet:
@frozenfas noticed an inconsistent unit in the help message of the defocus cli.
This sets it to the correct one instead (`nm` to `μm`)

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -713,9 +713,9 @@ def match_template(argv=None):
         required=False,
         action=ParseDefocus,
         help="Here you can provide an IMOD defocus (.defocus) file (version 2 or 3) "
-        ", a text (.txt) file with a single defocus value per line (in nm), "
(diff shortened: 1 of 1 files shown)


sroet:
@McHaillet  would 2d5979a actually work with an imod defocus file? (don't know their standards by heart)

McHaillet:
> @McHaillet would [2d5979a](https://github.com/SBC-Utrecht/pytom-match-pick/commit/2d5979af3bc908c443ba9f9705f71f8d7b212167) actually work with an imod defocus file? (don't know their standards by heart)

Indeed not, but the imod parsing has a separate function. The text file parsing is for some user made defocus file. Just for convenience if someone wants to parse data from for example CTFFIND to a text file input.

McHaillet:
Does that make sense to you to have it like this?

sroet:
> Does that make sense to you to have it like this?

Definitely!

McHaillet (approved):
👍

PR #199, opened 2024-08-02
mkdocs site for pytom-match-pick

McHaillet:
The site looks good to me locally, so time to make a PR!

Some points about the workflow for site deployment:

* The workflow that I made hopefully works, but you have more experience with those so any suggestions are welcome. 
* I added a  conditional that should only trigger on the main branch of the org's repository, and not on main of forks, is that needed or would it also work without that?


docs/Usage.md
```diff
@@ -0,0 +1,216 @@
...
+code = ("""
+---8<--- "entry_points.py:create_template_usage"
+""")
+cleaned_lines = [line.lstrip() for line in code.splitlines()
+    if not any(keyword in line for keyword in ('action=', 'default=', 'type='))
+]
+cleaned_code = '\n'.join(cleaned_lines)
+exec(cleaned_code)
+print(parser.format_help())
+```
```
sroet, line 41: I assume this is some mkddoc magic?

docs/Usage.md
```diff
@@ -0,0 +1,216 @@
...
+  angles used to reconstruct the tomogram. You should then also set the 
+  `--per-tilt-weighting` flag.
+* CTF data: a `.defocus` file from IMOD or `.txt` file to `--defocus-file`. The `.
+  txt` file 
+  should specify the defocus of each tilt in $\mu m$. You can also give a 
+  single defocus value (in $\mu m$). The CTF will also require input for 
+  `--voltage`, `--amplitude-contrast`, and `--spherical-abberation`.
+* Dose weighting: a `.txt` file to `--dose-accumulation` with the accumulated dose per 
+  tilt (assuming the same ordering as `.tlt`). Each line contains a single float 
+  specifying the accumulated dose in $e^{-}/\AA^{2}$. Dose weighting only works in
```
sroet, line 92: This does not render correctly for me in the diff screen (probably because [\AA was not-supported in mathJax](https://github.com/mathjax/MathJax/issues/795), but that should have been fixed in [this PR](https://github.com/mathjax/MathJax-src/pull/923/files)
```suggestion
  specifying the accumulated dose in $e^{-}/\unicode{212B}^{2}$. Dose weighting only works in 
```

docs/Usage.md
```diff
@@ -0,0 +1,216 @@
...
+cleaned_code = '\n'.join(cleaned_lines)
+exec(cleaned_code)
+print(parser.format_help())
+```
+
+## 3. Extracting particles
+
+Both scripts run on the job file created in `pytom_match_template.py` which contains details about correlation statistics and the output files. The job file always has the format `[TOMO_ID]_job.json`.
+
+**IMPORTANT** For both scripts the `[-r, --radius-px]` option needs to be considered carefully. The particle extraction will mask out spheres with this radius around each peak in the score volume and prevents selecting the same macromolecule twice. It is specified as an integer **number of pixels** (not Angstrom!) and ideally it should be the radius of the particle of interest. It can be found by dividing the particle radius by the pixel size, e.g. a ribosome (r = 290A / 2) in a 15A tomogram should gets a pixel radius of 9.6. As it needs to be an integer value and ribosomes are not perfect spheres, it is best to round it down to 9 pixels.
```
sroet, line 128: use angstrom sign
```suggestion
**IMPORTANT** For both scripts the `[-r, --radius-px]` option needs to be considered carefully. The particle extraction will mask out spheres with this radius around each peak in the score volume and prevents selecting the same macromolecule twice. It is specified as an integer **number of pixels** (not Angstrom!) and ideally it should be the radius of the particle of interest. It can be found by dividing the particle radius by the pixel size, e.g. a ribosome (r = 290Å / 2) in a 15Å tomogram should gets a pixel radius of 9.6. As it needs to be an integer value and ribosomes are not perfect spheres, it is best to round it down to 9 pixels. 
```

sroet (changes requested):
Couple comments, and please make sure all the images are actually needed for the tutorial, LGTM otherwise

docs/Usage.md
McHaillet: This requires markdown-exec, the workflow file also installs it.
That reminds me, I should also add  mkdocs, mkdocs-material, and markdown-exec to the dev  install dependencies.

docs/Usage.md
sroet: seems like those are also not allowed in github markdown...
But from some testing `\text{Å}` is :smile:

McHaillet: that works!

docs/Usage.md
McHaillet: added those dependencies to the pyproject.toml

McHaillet:
> Couple comments, and please make sure all the images are actually needed for the tutorial, LGTM otherwise

Unused files have also been removed!

sroet (approved):
Thanks for the changes, this LGTM, feel free to merge

PR #201, opened 2024-08-02
Quick fix README and workflow

McHaillet:
Noticed some formatting problems in the project README due to the snippets.

Also the site deployment fails due to incorrect string symbols.


sroet:
Does this generate correct docs, or are the --> things included?

PR #204, opened 2024-08-02
update workflow permission to write github pages

McHaillet:


McHaillet:
This has some info about deploying mkdocs: https://github.com/marketplace/actions/deploy-mkdocs

It says the following: 

>  You may need to give the GITHUB_TOKEN write permission. Go to your repository's Settings > Actions > General > Workflow Permissions and select Read and write permissions.

I am not sure whether permission contents: write does the same. As github docs says contents only controls this:

> Work with the contents of the repository. For example, contents: read permits an action to list the commits, and contents: write allows the action to create a release. For more information, see "[Permissions required for GitHub Apps](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents)."

McHaillet:
Should look here perhaps: https://stackoverflow.com/questions/72851548/permission-denied-to-github-actionsbot

McHaillet:
@sroet I think you were right and it should be "contents: write"

sroet (approved):
LGTM! 🤞 Let's hope this is enough to get it up and running

PR #205, opened 2024-08-02
Update Developers.md

McHaillet:
I read that to launch the site for the first time you need to manually push to the gh-pages branch. I did that and it is now live! ([sbc-utrecht.github.io/pytom-match-pick/](https://sbc-utrecht.github.io/pytom-match-pick/))

This PR is to check whether the docs indeed get updated now with a merge to main. (I corrected some poor sentences on one of the pages)


sroet (approved):
LGTM

PR #210, opened 2024-08-07
add check on tomo mask shape

### LINKED ISSUE #194: --tomogram-mask option during match_template not working
Hi again,
When I look at my score maps I see peaks inside the area where my tomogram mask should have zero values. I assume this shouldn't be the case if those subvolumes are skipped. 

My mask is from Amira and has the right dimensions. However, the pixel size is 1.00. Now I was wondering if the pixel size of the mask needs to match the tomogram? 
Or there is another problem at play :)

Thanks a lot!

sroet:
closes #194 
(the part that the tm-job should error out if the mask is the wrong size)

Also adds the same check to the extraction job

- [X] add checks to catch mismatch between tomo mask and tomogram in tmjob
- [x] as above for extraction
- [x] alter tooltips to mention number of pixels
- [x] add tests

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -436,7 +436,7 @@ def extract_candidates(argv=None):
         type=pathlib.Path,
         required=False,
         action=CheckFileExists,
-        help="Here you can provide a mask for the extraction with dimensions equal to "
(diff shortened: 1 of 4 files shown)


McHaillet (approved):
Thanks for the fix. LGTM!

PR #219, opened 2024-08-14
add comment to entry_points that tells developers not to use line bre…

McHaillet:
…aks in strings in the snippet


sroet (approved):
LGTM, feel free to merge

Issue #214, opened 2024-08-08
Extracting subtomos in RELION5 with particles.star file not working


Phaips:
Hello,

I wanted to extract subtomos in RELION5 (beta-3-commit-b978db) using a .star file from TM with `--relion5-compat`.
I run the Extract job from GUI but the command is as follows:
``which relion_tomo_subtomo_mpi` --p ../../rbcPipeline/pytom/rca/particles.star --t Tomograms/job015/tomograms.star --theme classic --o Extract/job053/ --b 32 --bin 4 --min_frames 1 --float16  --stack2d  --only_do_unfinished  --j 4  --pipeline_control Extract/job053/`

and the error I get is:
```
in: /scicore/projects/scicore-p-structsoft/ubuntu/software/RELION/ver5.0/src/jaz/tomography/particle_set.cpp, line 19
ERROR: there are no particles in ../../rbcPipeline/pytom/rca/particles.star
```

I am not sure why it does not recognise any particle, the .star file looks fine to me:
![image](https://github.com/user-attachments/assets/88b60320-ba9c-4125-bfe5-c2b4c260ed34)

I also tried giving absolute paths, only using a single particle list (not a merged one), also using reconstructed tomos in halfsets, etc. but it's always the same error so assume its a problem with the .star file. Maybe this is a RELION5 issues but I thought I anyway ask here first :)
Thanks for the help!

All the best,
Philippe

McHaillet:
Hi Philippe, 

Thanks for the feedback and solving the `data_particles` header! That definitely needs to be fixed.

- [ ] change star header to `data_particles` instead of just `data_`

I wrote the headers based on the Relion5 biorxiv paper and assumed they would work, but I didn't have a place to test the integration. 

Could you try changing the _rlnDetectorPixelSize to _rlnTomoTiltSeriesPixelSize and see if that makes it readable for Relion5?

> ```ParticleSet::ParticleSet: rlnTomoTiltSeriesPixelSize missing from optics MetaDataTable```  

Although this could also mean that there needs to be an optics table above it... Not sure.

Best,
Marten

McHaillet:
To-do:

- [ ] change star header to `data_particles` instead of just `data_`

Phaips:
Thanks a lot for the quick fix!

McHaillet:
Thanks for the quick response, I am glad that it solved it!

What do you mean with the `rec_` in the tomo name? In the name of the mrc file?

[...]

PR #220, opened 2024-08-14
Fix relion5 compat

McHaillet:
Fixes some of the problems addressed in #214 .

* starfiles are now written with `data_particles` instead of `data_` by providing a block name to the starfile package
* a column name in relion5 compat mode has been updated from `rlnDetectorPixelSize` to `rlnTomoTiltSeriesPixelSize`

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -548,7 +548,9 @@ def extract_candidates(argv=None):
 
     # write out as a RELION type starfile
     starfile.write(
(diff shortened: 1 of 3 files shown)


sroet (approved):
would these starfile still be readable in RELION 4 (as RELION 5 is technically still in public beta)? LGTM otherwise

McHaillet:
Yes, already from 3.1 they changed to the `data_particles` notation, but I guess were still excepting just `data_` (https://relion.readthedocs.io/en/release-4.0/Whats-new.html#release-3-1). So this is a safe fix for compatibility with the older version as well.

PR #222, opened 2024-08-14
Enable 16 bit output

### LINKED ISSUE #162: save score volume as float16
The score volumes from template matching could potentially be saved in MRC file mode 12 (float16) to save some disk space. (this was a recent update in the MRC mode)

sroet:
closes #162

- [x] add code
- [x] test added code


McHaillet (commented):
I had a quick look and before I review in detail, I was wondering if saving the angle index volume in float16 works for a large rotation search space. The angular search can be on the order of a million values and the integer precision needs to be high enough to store those indices. Does that still work with float16?

sroet:
> I had a quick look and before I review in detail, I was wondering if saving the angle index volume in float16 works for a large rotation search space. The angular search can be on the order of a million values and the integer precision needs to be high enough to store those indices. Does that still work with float16?

Yeah, you're right. Max is ~65000. Will change that back tomorrow

sroet:
@McHaillet thanks for catching that, changed it back. Please have another look

McHaillet (approved):
LGTM! Also nice that you added a test for writing an integer volume to MRC.

PR #223, opened 2024-08-14
logging.warn will be deprecated

McHaillet:
Need to change logging.warn to logging.warning as the former will be deprecated.


sroet (approved):
LGTM, feel free to merge

PR #224, opened 2024-08-15
deal with new ruff version and next try for tutorial-test.yml

sroet:
this is the next try at testing the tutorial, will now only trigger on pushes to main.
(will have to merged before it triggers)

might solve #148 

Also tries to fix the ruff issues with the added ipynb ([a new ruff version has released with ipynb linting enabled](https://github.com/astral-sh/ruff/releases/tag/0.6.0)) and updates the pre-commit hook

Also updates minimal coverage to 87%


docs/benchmarks/defocus_gradient_analysis.ipynb
```diff
@@ -13,8 +13,8 @@
     "import numpy as np\n",
     "from scipy.spatial.distance import cdist\n",
     "from scipy.optimize import curve_fit\n",
-    "from scipy.stats import t\n",
-    "sns.set(style='ticks', context='notebook')"
+    "\n",
```
McHaillet, line 16: Do all the double quotes also need to be escaped?

McHaillet: Nvm, they are inside double quotes

sroet: yeah, the double quotes and \n is how an ipynb looks like outside of jupyter

McHaillet (approved):
LGTM and anyway need to try the workflow right

PR #225, opened 2024-08-16
Update tutorial-tests.yml

sroet:
Next try at #148 

apparently ever run step is spawned independently.
This should set the correct working directory


McHaillet (approved):
I understand, looks like the right fix

PR #226, opened 2024-08-16
Get tutorial-tests.yml working

### LINKED ISSUE #148: Test the tutorial

sroet:
Alright, I am rerunning this job only on this branch until I get it working

closes #148

### DIFF
--- a/.github/workflows/tutorial-tests.yml
+++ b/.github/workflows/tutorial-tests.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies, code, and list everything
         working-directory: ./
         run: |
-          conda install -y -c conda-forge python=3 cupy cuda-version=11.8 curl
+          conda install -y -c conda-forge python=3 cupy cuda-version=11.8 curl unzip
           python -m pip install coverage mdextractor #mdextractor is new and might need to be replaced later
           python -m pip install .[dev]
           conda list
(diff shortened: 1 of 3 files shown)


sroet:
@McHaillet This finally works! It runs in 101 min, and can be seen here: https://github.com/SBC-Utrecht/pytom-match-pick/actions/runs/10425169098/job/28875465809 (commit 4ba5eda, which failed on ruff stuff in the regular unitests)

With that runtime, it might be worthwhile to actually pip install cupy, what do you think?

sroet:
It also already found the first error in the tutorial :sweat_smile:

McHaillet (approved):
LGTM

McHaillet:
> It also already found the first error in the tutorial 😅

Nice! What was the error?

sroet:
> > It also already found the first error in the tutorial 😅
> 
> Nice! What was the error?

the file you get from emdb is called  `emd_2938.map` instead of `emd_2938.mrc` (was mentioned that way in the file structure)

PR #228, opened 2024-08-29
add check on input gpu indices

### LINKED ISSUE #99: template matching might fail silently when unavailable gpu's are specified
As title, and I am getting some reports that it might actually fail silently if users try to run the commands with `-g 0 1 2 3`

sroet:
closes #99 by failing loudly

- [X] implement check for parser
- [x] add test


McHaillet (approved):
That all looks solid to me, feel free to merge!

PR #230, opened 2024-09-27
Drop py 39 and upgrade

sroet:
following [SPEC0](https://scientific-python.org/specs/spec-0000/)/[NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) we should probably drop python 3.9.

Also does some upgrading by running `ruff check --extend-select=UP` and either doing the auto-fix or manually fixing  the last couple


McHaillet (approved):
LGTM!

Issue #236, opened 2024-11-06
release failing


McHaillet:
The release for 0.7.4 is failing. The logs mention some [missing file](https://github.com/SBC-Utrecht/pytom-match-pick/actions/runs/11642425126/job/32627129353).

Seems similar to what was reported here: https://github.com/pypa/gh-action-pypi-publish/issues/291

McHaillet:
Is there a way to revert the commits on the main branch?

sroet:
Not easily, and it is fine for now. I will make an issue on their tracker

McHaillet:
The action was updated to v1.12 last week: https://github.com/pypa/gh-action-pypi-publish/releases/tag/v1.12.0

Looking at the quirks in the release notes it mentions something about self-hosted runners. That does coincide with the timeframe when our release started failing. Maybe we should downgrade to v1.11 instead.

McHaillet:
And report an issue

sroet:
yeah, let's pin to 1.11 until that issue has been resolved

McHaillet:
Right, at least that fixed the part where the action docker file could not be found, however now this happens:

https://github.com/SBC-Utrecht/pytom-match-pick/actions/runs/11781022041/job/32812923743

Actually, I recall now that this what I initially had when I tried to release 0.7.4, but then I reran deployment and got the other error. So now I am fairly confused

McHaillet:
OMG attestations are mentioned in the v1.11 update: https://github.com/SBC-Utrecht/pytom-match-pick/actions/runs/11781022041/job/32812923743

This is not really Okham's razor, but what I think might have happened is that I released 0.7.4 when the action was on v1.11 (while 0.7.3 was on v1.10), this caused the attestation problem that we see now. I did not immediately click redeploy and v1.12 might have come out in between the initial 0.7.4 deployment and the time I tried to rerun the workflow, hence we saw the second error.  Alternatively, I might have gone crazy lol

McHaillet:
Seems to have something to do with this: https://docs.pypi.org/attestations/producing-attestations/

Apparently we need to Trusted publishing to use attestations. Need to read into it more.

McHaillet:
As a workaround we know use v1.11 while setting attestations to False, as in PR #240. We should upload attestations though.

McHaillet:
[...]

PR #237, opened 2024-11-11
test if version update fixes release to test-pypi

McHaillet:
For issue #236 

I am unsure if this would solve it, but I do not know how else to diagnose it. As said in the discussion the issue sounded very similar to another reported issue. Perhaps this could fix it.

I only changed the version of the action for the test-pypi release here, as this is a test.


sroet:
please also update the version number (so we can cut another release) but otherwise looks good to me! fingers crossed

sroet (approved):

McHaillet:
done!

PR #238, opened 2024-11-11
set to 1.12 instead of 1.12.1

McHaillet:
I think I made the mistake of pointing it to 1.12.1 while it can only point to branches of the github action repo. Therefore 1.12 should be valid.


sroet:
So I don't think this works as all of those tags are identical to the failing one `v1`, but let's see.

sroet (approved):

McHaillet:
You're right, but maybe rerunning after the update of pypa works (that was 5 days ago). But then the v1 should also work, let's see

PR #240, opened 2024-11-11
dont publish attestation to pypi

McHaillet:
This should finally fix it then.


sroet (approved):
I think this is ok for now. Ideally we should upload attestations, but getting releases to work at all has priority

PR #241, opened 2024-11-15
[DO NOT MERGE] debug pypi workflow

sroet:


sroet:
@webknjaz, current results from running the [`ls` commands](https://github.com/SBC-Utrecht/pytom-match-pick/actions/runs/11857330471/job/33045462198?pr=241#step:7:1) inside the container:
`ls /home` returns nothing
`ls /__w/_actions` returns
```
actions
pypa
```
anything else you want me to check?

webknjaz:
But what about the subdirs? `ls - R` should show that

webknjaz:
So it seems to me like this is a case of a bug in GitHub Runner. And my hunch is that it's probably not caused by it being a hosted runner. It's likely caused by the use of the container setting on the job.

@sroet to verify this, drop the container from the job and try again.

If that's the case, I don't think we'll be attempting to support it (you can probably hack on your container and stick symlinks into it, but that's out of the scope).

Though, I also noticed that you're not following the recommendation of keeping the build invocation out of the job that has elevated OIDC privileges. Perhaps, if you just split it according to https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/, the problem will disappear.

sroet:
> @sroet to verify this, drop the container from the job and try again.

That's not happening as dropping the container from a self-hosted runner will run the code on the actual host machine. Our host machines are not isolated/revertible enough to let random code run on it outside of a container. 

> Perhaps, if you just split it according to https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/, the problem will disappear.

I can try that, and if that doesn't work maybe push the publishing part to github-runners instead of self-hosted ones

webknjaz:
> maybe push the publishing part to github-runners instead of self-hosted ones

That's exactly what I meant.

sroet:
So, splitting out the job to run the upload on github runners instead of self-hosted ones seems to work (the failure is expected as this branch can't be valid for the trusted publishing setup we have on this project)

sroet:
I think all the necessary debugging has been done, so I will close this. 
@webknjaz thanks for the help and please let me know if you want to test some other things! (Otherwise I will delete this branch in a week)

PR #250, opened 2025-01-08
improve center approximation for relion5

McHaillet:


sroet (changes requested):
Hmm, this feels wrong to me, as this is arguably worse for unbinned tomograms (but better for any form of binning).

For now I agree that this is the correct solution, but I would like to have an explicit comment added why we don't do the exact result and we should probably see if we can have a better estimate if we have relion 5 input data as well (feel free to turn that into an issue)

src/pytom_tm/extract.py
```diff
@@ -348,7 +348,7 @@ def extract_particles(
 
     if relion5_compat:
         dims = np.array(job.tomo_shape)
-        center = dims / 2 - 1
+        center = dims / 2
```
McHaillet, line 351:
```suggestion
        center = dims / 2  # we approximate the center, the correct way would be to use the
        # original dimensions and calculate: center = (ubinned_dims / 2) - 1 (as is done in relion5)
```

McHaillet:
> Hmm, this feels wrong to me, as this is arguably worse for unbinned tomograms (but better for any form of binning).
> 
> For now I agree that this is the correct solution, but I would like to have an explicit comment added why we don't do the exact result and we should probably see if we can have a better estimate if we have relion5 input data as well (feel free to turn that into an issue)

I fully agree with what you are saying, this should be treated as a workaround. But there might be situations where a tomogram is reconstructed outside relion and coordinates need to be imported to relion, for that situation this is probably the best we can do.

PR #255, opened 2025-01-21
new ruff version

sroet:
Currently our unit tests are failing because ruff changed some formatting rules around. This updates our files to be ruff compliant. Also updates a typo ('where asked' `->` 'were asked')


McHaillet (approved):
LGTM! Thanks for the fix :)

PR #264, opened 2025-02-20
Change extract radius to particle diameter for consistency

### LINKED ISSUE #262: Inconsistency between options in pytom_match_template.py and pytom_extract_candidates.py
Hi,

Apologies if I'm missing something obvious and I know this might be a very small issue, but I can't help:

Why does `pytom_match_template.py` has `--particle-diameter` in Å while `pytom_extract_candidates.py` has `--radius-px`?

Why not:
* Ask for either the diameter or the radius in both
* Set the units to either Å or px in both
* Have the same option name in both

I have a slight preference for using `--particle-diameter` in Å in both, as the diameter is what we usually measure directly from the tomograms, and keeping physical units avoids the need to do conversions at different binnings.

Thanks a lot!

McHaillet:
This makes the extract radius parameter easier to interpret and prevents redundancy. Closes #262

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -150,8 +150,8 @@ def predict_tophat_mask(
 
 def extract_particles(
     job: TMJob,
-    particle_radius_px: int,
     n_particles: int,
+    particle_diameter: float | None = None,
     cut_off: float | None = None,
     n_false_positives: float = 1.0,
     tomogram_mask_path: pathlib.Path | None = None,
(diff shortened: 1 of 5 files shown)


src/pytom_tm/extract.py
```diff
@@ -250,6 +250,22 @@ def extract_particles(
...
+    if particle_diameter is not None:
+        particle_radius_px = int((particle_diameter / 2) / job.voxel_size)
+    elif job.particle_diameter is not None:
+        particle_radius_px = int((job.particle_diameter / 2) / job.voxel_size)
+        logging.info(
+            "No particle diameter was provided, so using the diameter "
+            "specified previously to mask out areas around peaks. Take care for "
+            "strongly elongated particles as it might prevent correct "
+            "annotation when they arrange parallel to each other and close together."
+        )
```
sroet, line 262: This elif is not hit by the unit-tests, please alter one of the tests to grab the particle diameter from the previously run job.

sroet (changes requested):
One part missed by unittests, LGTM otherwise

sroet (approved):
LGTM, feel free to merge (commit came over cleanly)

McHaillet:
Great!

PR #270, opened 2025-03-04
Fix handling of tomogram masks in signed bytes

rdrighetto:
Ouch, I renamed the branch to something more meaningful and lost the old PR (???), here is it again, this time in conformity with pre-commit hooks 🙂


rdrighetto:
Hi @sroet, thanks for the feedback!
As for the test, I need to think how best to implement it. We do get tomogram masks as signed bytes sometimes with IMOD operations (not sure exactly what triggers this behavior) or from Amira. I could probably cook some python code to fake one.

You are right that ultimately, the `mrcfile` package should be able to handle this and we should not have to worry about it. But I have no idea how to implement it there 😅 
I can maybe open an issue and see how they respond to it?

src/pytom_tm/extract.py
```diff
@@ -247,6 +247,12 @@ def extract_particles(
             for origin, size in zip(job.search_origin, job.search_size)
         ]
         tomogram_mask = tomogram_mask[*slices]
+
+        # If tomogram_mask is stored as signed bytes we need to convert it to unsigned:
+        if tomogram_mask.dtype == "int8":
+            if np.any(tomogram_mask < 0):
+                tomogram_mask = (tomogram_mask > -128).astype("int8")
```
sroet, line 254: I would join these statements and make them strict to only trigger on misread signed binary masks
```suggestion
        if (tomogram_mask.dtype == "int8" and 
            -128 <= tomogram_mask.min() and tomogram_mask.max() <= -127
            ):
            tomogram_mask = (tomogram_mask > -128).astype("int8")
```

sroet (changes requested):
Thanks, looks good in principle, but as mentioned before, the altered lines are not hit by the test. Also, I think we want to be very strict when this behavior is triggered.

Most of the tests for the extraction are done in `tests/test_tmjob.py`
The test masks are written in the `setUpClass` method of the `TestTMJob` class and then used in the `test_extractions` method

McHaillet:
@rdrighetto, should we continue this? I can also write a test if needed :)

rdrighetto:
Sorry for the very slow response. Please let me try giving a shot at it, I'll let you know otherwise. Thanks for the offer!

McHaillet:
> Sorry for the very slow response. Please let me try giving a shot at it, I'll let you know otherwise. Thanks for the offer!

Sounds good!

sroet (approved):
Hey @rdrighetto, thanks for the fix and the test! Looks great, merging!

rdrighetto:
Thanks @sroet! Just to keep things properly linked, this fixes the problem described in [this comment](https://github.com/SBC-Utrecht/pytom-match-pick/discussions/198#discussioncomment-12375340) under #198

PR #274, opened 2025-03-10
update help on particle-diameter to describe intended behavior

### LINKED ISSUE #273: particle-diameter cutoff does not correspond to diameter
Hello,

When I extract my particles and I give the "new" particle-diameter now and not the radius in pixels anymore, it gives a wrong distance threshold for my particles. This is independent whether I overwrite this flag in the extract job or use the one from the matching job.
When I give 140 A my "lowest" particle distances are 77A and when I give 280 A my lowest particles distances are 144 A as expected. So somehow it gets divided once too much - because now the center to center distance is the radius which should not be the case right? 

Cheers,
Philippe

sroet:
Describe the behavior that the minimal peak-to-peak distance will be diameter/2 instead of the (apparently) expected full diameter

closes #273

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -324,6 +324,7 @@ def estimate_roc(argv=None):
         action=LargerThanZero,
         help="Particle diameter of the template in Angstrom. It is used during "
         "extraction to remove areas around peaks to prevent double extraction. "
+        "Minimal peak-to-peak distance after extraction will be diameter/2."
         "If not previously specified, this option is required. If "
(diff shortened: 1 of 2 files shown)


sroet:
I don't think this has a rush so it might be good to let @FridoF do it. However, I am pretty sure he won't have time to look at it in the coming weeks/months due to his currently increased workload.
If it is ok for you to leave it hanging for a while then that's ok.

McHaillet:
Okay, in that case I dont mind doing it!

McHaillet (approved):
LGTM!

PR #276, opened 2025-04-04
Deal with multi column tilt file

### LINKED ISSUE #275: multiple columns in rawtlt.txt crash file reader
Ioannis reported this: if the rawtlt.txt has a second column of values, read_txt_file() will crash. I would expect the .rawtlt file to only have a single value but I guess it is not really a defined file format. Maybe easy fix to only read the first value of a line for the .rawtlt. 

@sroet maybe easy fix?

sroet:
closes #275


sroet:
Alright, this is ready for review. It errors out on the regular input, but I added an `--tilt-angles-first-column` replacement for `--tilt-angles` that reads the file and only takes the first column (with a warning).

Added tests for all new behavior code and updated the version to 0.8.1 (as I think this is added functionality without breaking previous api)

McHaillet (approved):
I went through it, looks solid to me!

PR #277, opened 2025-04-09
Make sure we error with a reasonable message if gpus can't evenly devide number of splits

### LINKED ISSUE #271: number of gpu's that is not integer divisible by number of subvolumes gives unhelpful error message
Saw this issues today. The tomo was split into 4 subvolumes with -s 2 2 1, with 4 and 8 gpu's it worked fine; with 8 the rotation search is correctly split; but with 5, 6, and 7 gpu's it all produced the same incomprehensible error. 

In principal it is good that it doesn't work with 5 gpu's and and 4 subvolumes, because it will lead to situation where certain gpu's will be idle. Not ideal for cluster jobs. But, it would be helpful if the program raises a ValueError with a clear message in these situations.

sroet:
closes #271

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -44,7 +44,7 @@ jobs:
           path: coverage.xml
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           pull_request_number: ${{ steps.get-pr.outputs.PR }}
-          minimum_coverage: 88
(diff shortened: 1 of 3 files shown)


sroet:
This is ready for review, added a reasonable error and test for that error (and bumped minimal coverage)

McHaillet (approved):
Thanks for the update! LGTM

PR #283, opened 2025-05-09
informative error on mask template shape mismatch

### LINKED ISSUE #282: Give informative error if template and mask do not have the same dimensions
See #281 for the current deep error we get, this should be caught (earlier) and give a more informative error message

sroet:
closes #282


tests/test_tmjob.py
```diff
@@ -255,6 +271,23 @@ def test_tm_job_errors(self):
                 voxel_size=1.0,
                 tomogram_mask=TEST_WRONG_SIZE_TOMO_MASK,
             )
+        # Test template mask mismatch
+        with self.assertRaisesRegex(
+            ValueError,
+            rf"{TEMPLATE_SIZE, TEMPLATE_SIZE, TEMPLATE_SIZE}"
```
McHaillet, line 277: Not really familiar with these rf"" strings - I am just curious: what does it do and why is it handy in this case?

McHaillet (approved):
Sorry, forgot to review
LGTM!

tests/test_tmjob.py
sroet: They are a combination of f-string and raw-string, which helps with all the strange symbols that can show up in regex strings. Don't think they are technically needed here

McHaillet: thanks!

PR #284, opened 2025-05-09
make pytom warp xml compatible

shahpnmlab:
This commit provides the user with an additional argument to use warptools xml data to generate an accurate PSF model for convolution with the template prior to matching. I tested this modification with my dataset I can see that i was able to generate a CC map that accurately localises my protein of interest in the tomogram. 
image

The command I used was `pytom_match_template.py -t test_pytom/tmpl_5.8.mrc -v tiltseries/reconstruction/gs04_ts_003_5.80Apx.mrc -g 0  -m test_pytom/mask_5.8.mrc --warp-xml-file tiltseries/gs04_ts_003.xml --voxel-size-angstrom 5.8 --particle-diameter 130 --angular-search 12 --z-axis-rotational-symmetry 4 --random-phase-correction`
and this was the output to STDOUT
```
INFO:root:Will generate an angle list with a maximum increment of 12.0
INFO:root:Using an angle difference of 11.7265 for Z1 and X
INFO:root:Using an angle difference of 12.0000 for Z2
Progress job_0 on device 0:
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2550/2550 [02:14<00:00, 19.03it/s]
```
NOTES: 
1. I have hard-coded the defocus handedness to 0 (no correction) because the assumption is that the user would have run a handedness check before running template matching.
2. matching is performed on a per tomogram basis, perhaps you want to consider including a batch processing feature? In this case the ts basename can be accessed from either `mdocs/<ts_basename>.{mrc}.mdoc` or `tomostar/<ts_basename>.tomostar` dirs and the corresponding tomogram can be located in the <processing_dir>/reconstruction/<ts_basename>_%0.2f_Apx.mrc.


src/pytom_tm/entry_points.py
```diff
@@ -1008,6 +1017,16 @@ def match_template(argv=None):
...
+
+    elif args.warp_xml_file is not None:
+        voxel_size, tilt_angles, dose_accumulation, ctf_params, defocus_handedness = (
+            parse_warp_xml_data(
+                args.warp_xml_file,
+                args.tomogram,
+                phase_flip_correction=phase_flip_correction,
+            )
+        )
+        per_tilt_weighting = True
```
sroet, line 1029: I would not override the defocus handedness in the parse_warp_xml function, but instead set it to `args.defocus_handedness` inside this `if` statement (as is done on line 1037). This would keep the xml parser clean of random defaults that are not actually part of the xml, should default to `0` as that is the default for `args.defocus_handedness`, and still give a user the option to override the default if they absolutely want

src/pytom_tm/io.py
```diff
@@ -621,3 +621,96 @@ def parse_relion5_star_data(
         ctf_params,
         defocus_handedness,
     )
+
+
+def parse_warp_xml_data(
+        warp_xml_path: pathlib.Path,
+        tomogram_path: pathlib.Path,
+        phase_flip_correction: bool = False,
+        phase_shift: float = 0.0,
```
sroet, line 630: This is not used and directly overwritten on line 673

src/pytom_tm/io.py
```diff
@@ -621,3 +621,96 @@ def parse_relion5_star_data(
...
+    # First determine the tomogram_voxel_size from the tomogram_path
+    try:
+        # Try to read voxel size from MRC file
+        tomogram_meta = read_mrc_meta_data(tomogram_path)
+        tomogram_voxel_size = tomogram_meta["voxel_size"]
+    except Exception as e:
+        logging.warning(f"Could not read voxel size from tomogram: {e}")
+        # Default value if reading fails
+        tomogram_voxel_size = 1.0
+        logging.warning(f"Using default voxel size: {tomogram_voxel_size}. Results may be incorrect!")
```
sroet, line 659: This is a way too broad exception to catch and anything that throws this will kill the code later. It is probably best to not catch any error from this function, and just let it error out

src/pytom_tm/io.py
```diff
@@ -621,3 +621,96 @@ def parse_relion5_star_data(
...
+
+    def _flatten(t):
+        return [float(item) for sublist in t for item in sublist if item.strip()]
```
sroet, line 662: please move this function to right before you start using it (above lines 692-693)

src/pytom_tm/io.py
```diff
@@ -621,3 +621,96 @@ def parse_relion5_star_data(
...
+            "voltage": voltage * 1e3,
+            "spherical_aberration": spherical_aberration * 1e-3,
+            "flip_phase": phase_flip_correction,
+            "phase_shift_deg": phase_shift,
+        }
+        for defocus in defocus_values
+    ]
+
+    # Default defocus handedness to 0 (no correction) for compatibility with return value expectations
+    defocus_handedness = 0
```
sroet, line 708: As mentioned before, don't think this is needed inside this function

src/pytom_tm/io.py
```diff
@@ -621,3 +621,96 @@ def parse_relion5_star_data(
...
+
+    return (
+        tomogram_voxel_size,
+        flattened_tilt_angles,
+        flattened_tilt_dose,
+        ctf_params,
+        defocus_handedness,
```
sroet, line 715: as mentioned before, don't think this is needed

sroet (changes requested):
Thanks for the PR! mostly looks great, I did have a couple comments about the hard-coding of the defocus handedness.

> matching is performed on a per tomogram basis, perhaps you want to consider including a batch processing feature? 

As we run pytom-match-pick on a cluster with a scheduler, we normally run an adapted version of https://github.com/Phaips/batch_pytom_aretomo3 (that one loops over aretomo3 output, but could be altered to loop over warp output or just straight tomograms instead) or run the following script:
```bash
#!/bin/bash
 
# Define the directory containing the input files
INPUT_DIR="Tomograms/job017/tomograms/" # Adjust this path as needed
 
# Loop through files matching the pattern xaa, xab, etc.
for FILE in ${INPUT_DIR}*.mrc; do
    # Check if the file exists to avoid errors
    if [[ -f "$FILE" ]]; then
        echo "Submitting job for file: $FILE"
        # Submit the job with sbatch, passing the file as an argument
        sbatch --job-name=$(basename "$FILE") TMscript.sh "$FILE"
    else
        echo "No matching files found for pattern: xa*"
    fi
done
```
where TMscript.sh is
```bash
#!/bin/bash
#SBATCH --partition=gpuq                     # Queue (partition) name
#SBATCH --gres=gpu:1                         # Request 1 GPU
#SBATCH --output=job_%x_%j_output.log              # Standard output log
#SBATCH --error=job_%x_%j_error.log                # Standard error log
# Load necessary modules (if required)
# module load some_module
# Execute your command
pytom_match_template.py -t Human80SRibosomeTemplate.mrc -m mask.mrc -d tm_results_handednessInverted -v $1 --particle-diameter 300 --low-pass 35 --tomogram-ctf-model phase-flip --per-tilt-weighting --voxel-size 17.5 --random-phase-correction --relion5-tomograms-star Tomograms/job008/tomograms.star -s 2 2 2 -g 0
```

sroet:
Also, your code seems to not have run the pre-commit checks, please follow all the steps in the developer install section: https://github.com/SBC-Utrecht/pytom-match-pick?tab=readme-ov-file#developer-install

You can also by hand run `ruff check` to check the code and `ruff format` to auto-format the commit

McHaillet:
@shahpnmlab Thanks for moving it here Pranav!

For the precommit thing, you should be able to autoformat with this:

```
python -m pip install '.[dev]'
pre-commit install
pre-commit run src/pytom_tm/entry_points.py  # it looks like only this files has some strange tabs added!
```

Then just add and commit that file!

src/pytom_tm/io.py
shahpnmlab: Done

src/pytom_tm/io.py
shahpnmlab: Done

src/pytom_tm/io.py
shahpnmlab: Fair, i have removed this now

shahpnmlab:
> @shahpnmlab Thanks for moving it here Pranav!
> 
> For the precommit thing, you should be able to autoformat with this:
> 
> ```
> python -m pip install '.[dev]'
> pre-commit install
> pre-commit run src/pytom_tm/entry_points.py  # it looks like only this files has some strange tabs added!
> ```
> 
> Then just add and commit that file!

Thanks for the tip @McHaillet, all my ruff checks now pass. I am now jsut awaiting clarification on the returning of the values at the end of the function. Once I have that I will reinitate another pull request.

shahpnmlab:
Please can you review the changes here?
https://github.com/shahpnmlab/pytom-match-pick/blob/main/src/pytom_tm/io.py
https://github.com/shahpnmlab/pytom-match-pick/blob/main/src/pytom_tm/entry_points.py
(sorry still a Git newb)

src/pytom_tm/io.py
sroet: (currently on mobile, so sorry for any formatting) 

This was only for the return of the defocus_handedness variabel (GitHub shows a couple code lines before to help with context). It shouldn't be returned from this function, but instead be set inside the entry_point if statement.
Please let me know if this is still unclear!

shahpnmlab: Got it. I have now removed `defocus_handedness` from here and set it the value outside the `if` loop in the `entry_points.py`
I think I have resolved all out standing comments. Please let me know if the changes are sufficient :)

src/pytom_tm/entry_points.py
```diff
@@ -1008,6 +1019,18 @@ def match_template(argv=None):
...
+    elif args.warp_xml_file is not None:
+    voxel_size, tilt_angles, dose_accumulation, ctf_params = (
+        parse_warp_xml_data(
+            args.warp_xml_file,
+            args.tomogram,
+            phase_flip_correction=phase_flip_correction,
+        )
+    )
+    defocus_handedness = args.defocus_handedness
+    per_tilt_weighting = True
```
sroet, line 1032: This whole section needs to be indented
```suggestion
    elif args.warp_xml_file is not None:
        voxel_size, tilt_angles, dose_accumulation, ctf_params = (
            parse_warp_xml_data(
                args.warp_xml_file,
                args.tomogram,
                phase_flip_correction=phase_flip_correction,
            )
        )
    defocus_handedness = args.defocus_handedness
    per_tilt_weighting = True
```

sroet (changes requested):
Almost done! You're only now missing the indentation to only run the code when the elif is triggered. This code also has no test (yet), but I can add that once the base code is complete!

src/pytom_tm/entry_points.py
```diff
@@ -1008,6 +1019,18 @@ def match_template(argv=None):
...
+    elif args.warp_xml_file is not None:
+        voxel_size, tilt_angles, dose_accumulation, ctf_params = (
+            parse_warp_xml_data(
+                args.warp_xml_file,
+                args.tomogram,
+                phase_flip_correction=phase_flip_correction,
+            )
+        )
+    defocus_handedness = args.defocus_handedness
+    per_tilt_weighting = True
```
sroet, line 1032: I forgot to add these two in the indentation comment as well :man_facepalming: 
```suggestion
        defocus_handedness = args.defocus_handedness
        per_tilt_weighting = True
```

sroet: I will commit this one myself

shahpnmlab:
Sorry about that! I don't have an appropriate dev env on my machine (using the cluster GPUs) which is leading to these silly mistakes.

sroet:
> Sorry about that! I don't have an appropriate dev env on my machine (using the cluster GPUs) which is leading to these silly mistakes.

No worries! We are very happy with the addition of the code, and considering I can push to your branch to fix this, it is not too hard to work around. The final thing left is to add a test to cover the added code, which I will do. (The main reason is to make sure something notifies me if I break your code in the future). 
I don't think I will have time for that today, but probably work on it tomorrow! Thanks again for the addition :smile:

shahpnmlab:
Excellent! Happy to have contributed! The tool has definitely allowed increasing the fidelity of the picks as well as the speed. Being able to simply provide the xml file has reduced the need to write wrapper scripts to prepare inputs as well. I have been extolling the virtues of the tool here, so soon as the changes are merged, I will install the program in our software tree so that it is widely available (and used).

McHaillet:
Sorry had some busy days, will review on Monday

PR #296, opened 2025-06-25
actually use normalise function from correlation

sroet:
looking through our code to see if we can up our test coverage, starting with `correlation.py`

I noticed that we have a `normalise` function that does the same thing as was implemented in `matching.py`. Updated so that `normalise` is used.
Also dropped a `mask is not None` branch from `normalised_cross_correlation` function (this function is only used in testing, specifically `test_mask.py`)

I also updated the version to account for this API change, but advise against doing a release until some actual functionality is added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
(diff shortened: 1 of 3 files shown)


McHaillet (approved):
LGTM!

PR #297, opened 2025-06-25
add testing for create_mask

sroet:
This:
- adds tests for create mask entry point
- fixes a bug where if only `--radius-minor1` or `--radius-minor2` was given instead of both, we would silently create a spherical mask instead of raising an error
- add some tests for errors that can be raised from the `ellipsoidal_mask` function (making `mask.py` reach 100% 🥳 )
- rewrote test setup so that the output directory is removed and remade between tests (and uses temp directories)
- added a tests for the `_parse_argv`


McHaillet (approved):
LGTM, just one question and two grammar nitpicks, but feel free to merge

PR #300, opened 2025-07-18
add more tests

sroet:
This:
- [x] is PR 300 :partying_face: 
- [x] makes the tests for the masks more comprehensive (actually test if we can read the mrc's)
- [x] add a test helper `chdir` contextmanager that makes sure you can use `chdir` without breaking future tests if an error is raised during the `chdir` 
- [x] bumps minimal coverage to 91%
- [x] adds a secret '--log-test' option to the arguments to prevent losing the logging handlers during logtesting

### DIFF
--- a/tests/test_entry_points.py
+++ b/tests/test_entry_points.py
@@ -211,7 +211,7 @@ class TestEntryPoints(unittest.TestCase):
 
         # Test failure on trying to get more resolution
         args = defaults.copy()
-        args["--output-voxel-size-angstrom":"0.5"]
+        args["--output-voxel-size-angstrom"] = "0.5"
         with self.assertRaisesRegex(NotImplementedError, "smaller voxel size"):
             start(args)
 


tests/test_entry_points.py
```diff
@@ -148,14 +150,70 @@ def start(arg_dict):
...
+        args["--log-test"] = ""
+        with self.assertLogs(level="WARNING") as cm:
+            start(args)
+        self.assertEqual(len(cm.output), 1)
+        self.assertIn("voxel size does not match", cm.output[0])
+        self.assertTrue(output.exists())
+
+        # Test failure on trying to get more resolution
+        args = defaults.copy()
+        args["--output-voxel-size-angstrom"] = "0.5"
```
McHaillet, line 214: This made me think of that it might be nice if the map can also be upsampled, with the caveat that the maximal resolution is of course determined by the initial map. It would probably be relatively easy to add.

McHaillet (approved):
Nice updates! And great to see PR #300!

tests/test_entry_points.py
sroet: yeah, you can make an issue of this if you want, but I think it will become hard to track the "real" resolution. 

Also, most people can upsample a map in chimeraX if they really want to. (I am mostly worried about issues were we first upsample a template and then downsample again for matching, and there are a lot of upsampling algorithms with different caveats)

sroet:
Thanks for the review, merging

PR #303, opened 2025-07-29
make sure we can load a job.json from a different directory than it was run

### LINKED ISSUE #301: TMJob should save absolute paths for files
Just encountered an unexpected error when I ran `pytom_extract_candidate` from a different directory than the template_matching was run (I was inside the output directory instead), and it could not find `template.mrc`

sroet:
Should close #301 

This:
- [x] adds a test to see if we can load the json from the results directory, that fails on the current version of the code (mimicking issue #301, fails with the expected error on 35ffba1)
- [x] resolves the paths when writing to json
- [x] add a note that calling write_to_json from a different directory than the job was initiated breaks the paths
- [x] update version (I think this is a bugfix, not an API change)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.10.0"
+version = "0.10.1"
 description = "PyTOM's GPU template matching module as an independent package"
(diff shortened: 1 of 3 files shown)


McHaillet (approved):
awesome, very nice that this could be fixed by adding one pathlib call xD

PR #305, opened 2025-08-01
Keep track of relion5 binning and tilt series pixel size for better extraction

### LINKED ISSUE #251: fix coordinate centers with relion5 metadata input
For now I agree that this is the correct solution, but I would like to have an explicit comment added why we don't do the exact result and we should probably see if we can have a better estimate if we have relion 5 input data as well (feel free to turn that into an issue)

_Originally posted by @sroet in https://github.com/SBC-Utrecht/pytom-match-pick/pull/250#pullrequestreview-2537150941_

sroet:
closes #251 

This:

- [x] adds a `metadata` dictionary to TMJob to keep track of metadata that could be useful later on
- [x] uses the metadata to tag along the relion5 binning and tilt series pixelsize 
- [x] uses the new metadata to implement the correct center approximation mentioned as a comment in the code
- [x] updates the entry point docstring to mention the new behavior
- [x] add a function to testing_utilities to write out relion5 tomogram.star and tilt_series.star with the expected data

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -35,6 +35,8 @@ jobs:
         run: |
           ruff check
           ruff format --check
+      - name: Silence tqdm
+        run: echo "TQDM_DISABLE=1" >> $GITHUB_ENV
       - name: Run tests
         run: |
           coverage run -m unittest discover tests/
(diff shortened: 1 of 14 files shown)


sroet:
> internal metadata storing has become quite messy with all the accumulated changes (I should have put some better structures in place for it from the get go), but probably not worth the time investment to rewrite it

I think it would already help to make a tilt-series metadata class to keep track of that data if it exists. It is a bit of a rewrite, but not something we (or at least I) can't slowly work towards with new/updated implementations

sroet:
Also, thanks for the review

Issue #307, opened 2025-08-05
restructure metadata around TMJob


sroet:
internal metadata storing has become quite messy with all the accumulated changes (I should have put some better structures in place for it from the get go), but probably not worth the time investment to rewrite it

_Originally posted by @McHaillet in https://github.com/SBC-Utrecht/pytom-match-pick/pull/305#pullrequestreview-3087740202_

sroet:
I think the folowing improvements could be nice (list is not complete):
- [x] make tilt-series metadata use a dataclass and be fed to the algorithms via that
- [x] Group CTFData together (completed in #310)

McHaillet:
Agreed, a class that stores all the metadata could already do a lot. It will probably have a lot of optional attributes to be flexible for loading all the different metadata formats

Potentially the weighting functions could be attached to it (pseudocode):

```python
class TiltSeriesMetaData:
    init:
        self.tilt_angles
        self.defocus  # might be None
        self.dose  # might be None
        # etc...

    @property
    def template_weighting_function(self):
        # put create_wedge from src/pytom_tm/weights in here?
```

sroet:
Was there any other low hanging fruit other than `TomogramData` that is left?

PR #310, opened 2025-08-07
Structure CTF data in a CtfData dataclass

sroet:
begins solving #307 

- [x] write CtfData dataclass
- [x] make sure dataclasses can be saved/loaded in a generic way with the mixin-class and custom JSON en/decoders
- [x] write tests for dataclasses
- [x] rewrite code to use dataclasses
- [x] update docstrings of all the touched code

This also:
- [x] Adds tests for loading jobs of version 0.10.0 and earlier (also the new dataclass auto-fixes the old required 0.6.0 fix)

### DIFF
--- a/src/pytom_tm/dataclass.py
+++ b/src/pytom_tm/dataclass.py
@@ -28,27 +28,3 @@ class CtfData(JsonSerializable):
     spherical_aberration: float
     phase_shift_deg: float = 0.0
     flip_phase: bool = False
-
-
-# @dataclass(kw_only=True)
-# class TiltSeriesMetaData:
-#    tilt_angles: list[float]
-#    dose_accumulation: list[float] | None = None
-#    ctf_param: list[CtfData] | None = None
-#    per_tilt_weighting: bool = False
-#
-#    def __post_init__(self):
-#        """Check that if something is given it has the same length as tilt_angles"""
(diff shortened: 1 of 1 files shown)


sroet:
@McHaillet, normally I don't ping you on drafts, but it would be nice to have your opinion if the current grouping makes sense. Also, I am debating writing a tomogram dataclass as well (I am mostly struggling to find a place for the `voxel_size` that makes sense)

src/pytom_tm/dataclass.py
```diff
@@ -0,0 +1,30 @@
+from dataclasses import dataclass
+from pytom_tm.json import JsonSerializable
+
+
+@dataclass(kw_only=True)
```
McHaillet, line 5: neat with the kw_only to force the code to be more readable :)

src/pytom_tm/json.py
```diff
@@ -0,0 +1,49 @@
+from dataclasses import asdict, is_dataclass
+import json
+from typing import Any, Type
+
+
+class JsonSerializable:
```
McHaillet, line 6: I understand this is needed to make the custom CTF dataclass serializable, right? No built-in options available in dataclass?

McHaillet (approved):
Awesome, great change! Took quite some rewrite so good that you limited this PR to only ctf data

PR #311, opened 2025-08-20
Make tilt-series dataclass

sroet:
This PR:
- [x] makes a tilt-series metadata class
- [x] add documentation to the dataclass
- [x] alter all the code to use this class instead
  - [x] TMJob and downstream
  - [x] extract job and downstream
- [x] makes tilt_angle/tiltseriesdataclass not optional anywhere in the code
- [x] remove old seperated sanitation code to the dataclass
- [x] adds tests for all the new stuff
- [x] updates version number

### DIFF
--- a/tests/test_weights.py
+++ b/tests/test_weights.py
@@ -307,22 +307,6 @@ class TestWeights(unittest.TestCase):
             self.reduced_even_shape_3d,
             msg="Tilt weighted wedge should also work without defocus and dose info.",
         )
-        # TODO: this should be moved to a dataclass test
-        metadata = self.ts_metadata.replace(
-            dose_accumulation=None, ctf_data=CTF_PARAMS[:1]
-        )
-        weights = create_wedge(
-            self.volume_shape_even,
-            metadata,
-            self.voxel_size * 3,
-            per_tilt_weighting=True,
-            low_pass=self.low_pass,
-        )
-        self.assertEqual(
-            weights.shape,
-            self.reduced_even_shape_3d,
-            msg="Tilt weighted wedge should work with single defocus.",
-        )
 
     def test_ctf(self):
         ctf_data = CtfData(
--- a/tests/test_dataclass.py
+++ b/tests/test_dataclass.py
@@ -38,3 +38,12 @@ class TestTiltSeriesDataclass(unittest.TestCase):
             ValueError, r"same number of doses as tilt angles \(3\)"
         ):
             _ = temp.replace(tilt_angles=[1, 2, 3])
+
+    def test_ctf_expansion(self):
+        a = CtfData(
+            defocus=1, amplitude_contrast=0.1, voltage=300, spherical_aberration=1e-6
+        )
+        ts_metadata = TiltSeriesMetaData(tilt_angles=[1, 2, 3], ctf_data=[a])
+        self.assertEqual(len(ts_metadata.tilt_angles), len(ts_metadata.ctf_data))
+        for ctf in ts_metadata.ctf_data:
+            self.assertIs(ctf, a)


src/pytom_tm/dataclass.py
```diff
@@ -28,3 +28,116 @@ class CtfData(JsonSerializable):
...
     phase_shift_deg: float = 0.0
     flip_phase: bool = False
+
+    def replace(self, **kwargs):
+        return replace(self, **kwargs)
+
+
+@dataclass(kw_only=True)
+class TiltSeriesMetaData(JsonSerializable):
+    """A dataclass for to keep all the meta-data for the tilt-series together
```
McHaillet:
```suggestion
    """A dataclass to keep all tilt-series metadata together
```

src/pytom_tm/entry_points.py
```diff
@@ -1029,41 +1027,37 @@ def match_template(argv=None):
...
     if args.relion5_tomograms_star is not None:
-        (
-            voxel_size,
-            tilt_angles,
-            dose_accumulation,
-            ctf_params,
-            defocus_handedness,
-            rel_metadata,
-        ) = parse_relion5_star_data(
+        voxel_size, ts_metadata = parse_relion5_star_data(
```
McHaillet, line 1030: What is the reason for not combining the voxel size in the tilt-series metadata? I guess this is something we would rather assign to TomogramData once its added?

src/pytom_tm/weights.py
```diff
@@ -276,33 +274,24 @@ def create_wedge(
...
         low pass filter resolution in A
     high_pass: Optional[float], default None
         high pass filter resolution in A
-    tilt_weighting: bool, default False
-        apply tilt weighting
-    accumulated_dose_per_tilt: Optional[list[float, ...]], default None
-        accumulated dose for each tilt for dose weighting
-    ctf_params_per_tilt: Optional[list[CtfData, ...]], default None
-        ctf parameters for each tilt (see CtfData data class for specification)
+    per_tilt_weighting: bool | None, default None
```
McHaillet, line 287: Why the override? Is it needed for a specific case?

src/pytom_tm/weights.py
```diff
@@ -318,22 +307,21 @@ def create_wedge(
     elif cut_off_radius <= 0:
         raise ValueError("Invalid wedge cutoff: needs to be larger than 0")
 
-    if angles_in_degrees:
-        tilt_angles_rad = [np.deg2rad(w) for w in tilt_angles]
+    if ts_metadata.angles_in_degrees:
+        tilt_angles_rad = [np.deg2rad(w) for w in ts_metadata.tilt_angles]
```
McHaillet:
```suggestion
        tilt_angles_rad = np.deg2rad(ts_metadata.tilt_angles)
```
I know this is my code, but any reason we could not just do this?

tests/test_weights.py
```diff
@@ -128,54 +133,77 @@ def test_create_symmetric_wedge(self):
...
             self.volume_shape_even,
-            TILT_ANGLES,
-            1.0,
-            tilt_weighting=True,
-            ctf_params_per_tilt=CTF_PARAMS,
+            ts_metadata=self.ts_metadata,
+            voxel_size=1.0,
+            per_tilt_weighting=True,
+        )
+        sym_metadata = self.ts_metadata.replace(
```
McHaillet, line 166: This section is confusing me. This creates a binary wedge, right?
Why does it still have a list of two CTFs? And dose? For the binary wedge I thought it should be only one of each.

I get the implementation is not the best from my side, because theoretically you could have a tilt-series with two tilts and there is not way to distinguish it from the binary wedge input. Perhaps we could consider deprecating the binary wedge at some point.

McHaillet (commented):
Awesome work, this reads much cleaner and the subclassing of the metadata for Warp and Relion is really neat. This is my first pass with a few small comments and unclarities

src/pytom_tm/weights.py
sroet: The override is needed as the original code in tmjob.py always generates a binary wedge for the tomogram, and only uses per-tilt weighting on the template (thus requiring this override). Thinking about this I am not entirely sure that's correct...

src/pytom_tm/weights.py
sroet: nope, added.

tests/test_weights.py
sroet: Ah, that is because we directly call `create_wedge` which has some logic to determine if it should generate a symmetric wedge or not (by checking if the min/max angle are identical)
https://github.com/SBC-Utrecht/pytom-match-pick/blob/50ee18d48673a08f40861beea7da9d0ae641f6fa/src/pytom_tm/weights.py#L327-L338

and we tell users to input either 2 values or a file:
https://github.com/SBC-Utrecht/pytom-match-pick/blob/1ce3a27e4c96cd8b47c2d3b0db07653ef06faac5/src/pytom_tm/entry_points.py#L765-L768

(which we check for in `ParseTiltAngles` in `io.py`).

So a tilt-series metadata should never contain a single tilt (which we even check for in the following check:
https://github.com/SBC-Utrecht/pytom-match-pick/blob/50ee18d48673a08f40861beea7da9d0ae641f6fa/src/pytom_tm/dataclass.py#L73-L75

src/pytom_tm/entry_points.py
sroet: Yeah, a TomogramData class was what I was aiming for indeed, it felt "off" to add it to the tilt-series data as the tomogram can have different bining (as we see in Relion), at that point I would also probably move the Binning out of the RelionTiltSeries sub_class

sroet:
One other question I was thinking about, do we want to make the TiltSeriesDataClass frozen? as I don't think we can guarantee sanity if people don't use the replace. I will push a commit

src/pytom_tm/entry_points.py
McHaillet: Makes sense, agreed :+1:

src/pytom_tm/weights.py
McHaillet: In that case the additional switch is not needed anymore I guess.

tests/test_weights.py
McHaillet: I understood that, but I am confused why the dose and CTF are lists? Are the elements of those lists expected to be identical? And if they are not, which one is used?

src/pytom_tm/weights.py
sroet: The behavior is still identical (always generate binary wedge for the tomogram), I am just not certain that is correct. If we decide that is not correct, then we remove the additional switch

tests/test_weights.py
sroet: Ctf could be a single value (then it's copied for every tilt) Dose is not needed, so I could set that to `None` in the replace if we wanted, for the CTF it uses the following code for no-per-tilt weighted wedges:
https://github.com/SBC-Utrecht/pytom-match-pick/blob/1ce3a27e4c96cd8b47c2d3b0db07653ef06faac5/src/pytom_tm/weights.py#L351-L359

src/pytom_tm/weights.py
McHaillet: Discussed this a bit in person yesterday. I think we agreed (but please tell me if not the case) any weighting applied to the tomogram should preferably be binary, as otherwise it interferes in Fourier space with the interpolation and weighting of the tomographic reconstruction. It is an option to apply a fanned binary wedge to the tomogram to explicitly set everything to zero where we dont expect information relative to the template weighting, and that could be nice to add in the future. Might be as easy as:

```
fanned_binary_wedge = create_wedge(tilt_angles)
fanned_binary_wedge[fanned_binary_wedge != 0] = 1
```

tests/test_weights.py
McHaillet: Okay, that makes sense. Sounds good to me.

McHaillet:
> One other question I was thinking about, do we want to make the TiltSeriesDataClass frozen? as I don't think we can guarantee sanity if people don't use the replace. I will push a commit

Sounds good. We don't want the instance to be changed I think and this forces people to use replace if they would want to do that.

tests/test_weights.py
```diff
@@ -267,28 +294,29 @@ def test_create_wedge(self):
...
             low_pass=self.low_pass,
-            accumulated_dose_per_tilt=None,
-            ctf_params_per_tilt=None,
         )
         self.assertEqual(
             weights.shape,
             self.reduced_even_shape_3d,
             msg="Tilt weighted wedge should also work without defocus and dose info.",
         )
+        # TODO: this should be moved to a dataclass test
```
McHaillet: seems like this TODO was done?

sroet: Kinda, it is not testing any functionality at the moment as the dataclass will auto-populate the complete ctf list. I will try to remove this test and see if any lines are then untested

sroet: thanks for catching it, I moved it and it made no difference in test coverage

src/pytom_tm/weights.py
sroet: Made an issue #318 to keep track of this as I think that needs more in-depth testing

sroet:
Thanks again for the thorough review!

PR #313, opened 2025-09-02
Update CITATION.cff for zenodo

sroet:
This should fix the zenodo error.

Checked the old yaml via https://www.yamllint.com/ which gave a similar error as zenodo.
This fixes the yaml to become valid, and we should do a release after that.


McHaillet (approved):
Thanks for the fix!

PR #316, opened 2025-09-19
force tilt angles for TMJob

sroet:
This:

- [x] Forces a TMJob to get tilt angles
- [x] remove some unnecessary checks
- [x] make symmetric wedge behave with 90 degree tilts (0 degree wedge, used for testing)
- [x] Fix a bug in the behavior of volume splits
- [x] updates the version to represent the API break
- [x] make symmetric wedge error out for broken cases (negative angles or more than 90 degree angles)
- [x] make sure all touched code is tested

(Was a big rewrite that needed to be done as part of #311, so decided to split it out to a separate PR)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.11.1"
+version = "0.12.0"
 description = "PyTOM's GPU template matching module as an independent package"
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -1078,11 +1078,11 @@ def match_template(argv=None):
         args.template,
         args.mask,
         args.destination,
+        tilt_angles=tilt_angles,
         angle_increment=args.angular_search,
         mask_is_spherical=True
         if args.non_spherical_mask is None
         else (not args.non_spherical_mask),
-        tilt_angles=tilt_angles,
         tilt_weighting=per_tilt_weighting,
         search_x=args.search_x,
         search_y=args.search_y,
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -961,9 +961,10 @@ class TMJob:
             )
 
         # apply the optional band pass and whitening filter to the search region
-        search_volume = np.real(
-            irfftn(rfftn(search_volume) * tomo_filter, s=search_volume.shape)
-        )
+        # TODO: undo the debug comments
+        # search_volume = np.real(
+        #    irfftn(rfftn(search_volume) * tomo_filter, s=search_volume.shape)
+        # )
 
         # load rotation search
         angle_ids = list(range(self.start_slice, self.n_rotations, self.steps_slice))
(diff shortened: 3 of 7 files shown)


FridoF:
Hi both,

Since I do not have my cell phone with me for 2-factor authentication my response per email (will try to also put into system back home):


  *
The rationale for the filter was that reconstructions can actually have some signal in the missing wedge area - i.e., our iterative Fourier reconstruction did that (not always productive for template matching ...) and any modern AI reconstructions will try the same (also interpolating between the experimentally sampled data). It would be inconsistent to filter the template but leaving the variance outside the filter in for the tomogram. I agree that reconstructions from AreTomo probably do not require this filtering, but IsoNet and its many friends currently popping up will (or removing filter on the template side). If you want to remove, I would at least put a disclaimer that the wedge filtering should not be applied if you apply any kind of restoration.
  *
Yes, the filter would become an ellipsoid. It was relatively easy to deal with in real space, i.e., with the PSF (when placed into the centre of the image)

Best wishes

Friedrich






From: Marten Chaillet ***@***.***>
Date: Monday, 29 September 2025 at 11:02
To: SBC-Utrecht/pytom-match-pick ***@***.***>
Cc: Förster, F.G. (Friedrich) ***@***.***>, Mention ***@***.***>
Subject: Re: [SBC-Utrecht/pytom-match-pick] force tilt angles for TMJob (PR #316)

CAUTION: This email originated from outside of Utrecht University. Do not click links or open attachments unless you recognize the sender and know the content is safe.

[https://avatars.githubusercontent.com/u/58044494?s=20&v=4]McHaillet left a comment (SBC-Utrecht/pytom-match-pick#316)<https://github.com/SBC-Utrecht/pytom-match-pick/pull/316#issuecomment-3345806907>

To anyone reviewing this, please have a hard look at the change of behavior for split volumes. It used to be the case that we first subsliced the tomogram and then calculated and applied the filters on this subvolume, which I believe is wrong (as for example a nyquist filter does not make sense if we split the tomogram into singular pixels)

It now calculates and applies the tomogram filter on the complete tomogram before slicing down to the subvolume. For now this is done for every subvolume, but in the future we should probably rewrite the split volume function to pre-compute the filtered complete tomogram (this should become an issue after this PR)

About that, it might be better to remove the tomo_filter entirely. I copied this from the original pytom, but thinking it over it doesn't make much sense to me. There is two issues:

  *   The regions in the missing wedge should be zero for any sanely reconstructed tomogram, so we don't need to explicitly zero them.
  *   Generating a Fourier space filter for a non-square volume is not properly defined. I always assumed a non-square image/volume could be filtered with an ellipsoidal filter, but I don't think this actually holds. (Maybe ask Frido though?)

The second part is the bandpass. I don't think this needs to be applied to the tomogram either. Once it's applied to the template, the template does not have that information anymore, so it should already suffice for the search. Maybe also good to hear @FridoF<https://github.com/FridoF> opinion on this?

McHaillet:
Hi @FridoF, your mail is correctly forwarded to the thread in GitHub, so no worries!

Thanks for the input, the implementation of the ellipsoidal bandpass is correct in that case, some for wedge mask for the tomogram (which is also relative to the shape). Intuitively the ellipsoid makes sense to me, but just wasn't certain if it was valid.

@sroet I agree with your implementation to compute it on the full tomogram. It should also provide better sampling of the filters :)

McHaillet (approved):
Look like some great fixes to me :+1:

PR #324, opened 2025-11-24
use proposed fix for BetweenZeroAndOne argparse

### LINKED ISSUE #323: io.py class BetweenZeroAndOne(argparse.Action)
Hi,
This is my first "Issue" written here, so it's totally possible that I have it wrong. In io.py, I have a question about class BetweenZeroAndOne(argparse.Action), see below:

`

    class BetweenZeroAndOne(argparse.Action):
    """argparse.Action subclass to constrain an input value to a fraction, i.e. between
    0 and 1."""

        def __call__(
            self, parser, namespace, values: float, option_string: str | None = None
        ):
            if 1.0 <= values <= 0.0:
                parser.error(
                    f"{option_string} is a fraction and can only range between 0 and 1"
                )

            setattr(namespace, self.dest, values)
`

Specifically:

        `if 1.0 <= values <= 0.0:
            parser.error(
                f"{option_string} is a fraction and can only range between 0 and 1"
            )`

Is this evaluated as  (1.0 <= values) and (values <= 0.0)?

to get "True", both "1.0 <= values" and "values <= 0.0" must be true simultaneously?

values would have to be "at least 1.0" and "at most 0.0" at the same time? (not possible for any int, float, decimal, etc.?)

Would forcing values to be a fraction between 0 and 1 also be done with one of the two options? (I don't know which you prefer)

        `if not (0.0 <= values <= 1.0):
            parser.error(
                f"{option_string} is a fraction and can only range between 0 and 1"
            )`

or

        `if values < 0.0 or values > 1.0:
            parser.error(
                f"{option_string} is a fraction and can only range between 0 and 1"
            )`

All good if I didn't get this right :)

sroet:
This closes #323:

- [x] implement logic fix for the BetweenZeroAndOn argparse action as proposed in #323 
- [x] adds a test for the logic
- [x] updated version number for this bug fix

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.12.0"
+version = "0.12.1"
 description = "PyTOM's GPU template matching module as an independent package"
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/src/pytom_tm/io.py
+++ b/src/pytom_tm/io.py
@@ -107,7 +107,7 @@ class BetweenZeroAndOne(argparse.Action):
     def __call__(
         self, parser, namespace, values: float, option_string: str | None = None
     ):
-        if 1.0 <= values <= 0.0:
+        if not (0.0 <= values <= 1.0):
             parser.error(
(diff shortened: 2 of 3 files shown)


McHaillet (approved):
Great catch, LGTM

PR #327, opened 2025-12-15
Fix tutorial tests: deal with new dataverse download naming

sroet:
Our tutorial tests are failing in unzipping the `dataverse_files.zip` file as dataverse decided to rename the default data:

>  When there are no errors, this has the effect of saving the file under the name suggested by Dataverse (which as of v6.7 will be based on the persistent identifier of the dataset and the latest version number, for example doi-10.70122-fk2-n2xgbj_1.1.zip; in prior versions the file name was dataverse_files.zip in all cases).

([ref link](https://guides.dataverse.org/en/6.7.1/api/dataaccess.html))

This updates the unzip command to the new filename


McHaillet (approved):
Thanks for the fix!

sroet:
Still failing, but with the expected error (not using a dedicated test machine, so the GPU is in use)

PR #328, opened 2026-01-06
generate filters once for split jobs

### LINKED ISSUE #317: Only compute filters once with job splitting
It now calculates and applies the tomogram filter on the complete tomogram before slicing down to the subvolume. For now this is done for every subvolume, but in the future we should probably rewrite the split volume function to pre-compute the filtered complete tomogram (this should become an issue after this PR)

_Originally posted by @sroet in https://github.com/SBC-Utrecht/pytom-match-pick/issues/316#issuecomment-3338996334_

sroet:
This PR:

- [x] splits out filter generation to a separate method of TMJob
- [x] makes sure it is generated only once per split
- [x] add/update tests for everything
- [x] update version number
- [x] makes test coverage 100%

Some notes:
- The ctf updating and final wedge filters have to be generated per-split-job so I left that at the current spot
- We should probably make sure that if any of the relevant attributes are read only (or invalidate the cached filters), but I feel that will require a significant rethink about the structure of the TMJob class in general

closes #317

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.12.1"
+version = "0.12.2"
 description = "PyTOM's GPU template matching module as an independent package"
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -549,6 +549,22 @@ class TMJob:
         # output dtype
         self.output_dtype = output_dtype
 
+        # caching stuff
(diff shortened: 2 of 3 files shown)


sroet:
This is ready for review,
I split out all the filters that are not depended on the changes of the defocus value (which are different for every sub volume). I then made sure to cache the results and dealt with copying/json dumping)

I also decided to add some tests to bump the test coverage for the TMJob class (and the whole file) to 100% (from commit [5b6038c](https://github.com/SBC-Utrecht/pytom-match-pick/pull/328/commits/5b6038c11f83ff0af24bcf09295ce5a84f26e6e2) onward)

McHaillet (approved):
Great PR, looks clean to me!

PR #330, opened 2026-01-15
Ensure at least 2 pieces for GPU distribution check

McHaillet:
Add check for minimum pieces before volume split distribution check.

I was trying to split the rotation search of a single tomo over multiple GPU's (`-s 1 1 1` with `-g 0 1 2`). The current check doesn't allow for this apparently.

### DIFF
--- a/src/pytom_tm/parallel.py
+++ b/src/pytom_tm/parallel.py
@@ -87,7 +87,7 @@ def run_job_parallel(
     n_pieces = reduce(lambda x, y: x * y, volume_splits)
 
     # error out reasonably if we can' make an even distribution
-    if n_pieces % len(gpu_ids) != 0:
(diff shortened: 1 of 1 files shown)


sroet:
Do you mind adding a test to make sure the 1 GPU path now goes through? Otherwise LGTM!

McHaillet:
It might make more sense to move the check into the section for volume splitting actually. For angle splitting, multiple gpu's should be okay unless there is just one angle

sroet:
> For angle splitting, multiple gpu's should be okay unless there is just one angle

We could change that behavior to just check if there isn't more GPUs than angles in general, do need to make sure that this doesn't lead to double compute though (don't remember the splitting code by heart)

McHaillet:
We can close this, given the superior fix in #336

PR #334, opened 2026-01-21
change warp xml reader to multiply angles with -1

sroet:
From our own internal discussion, this seems like the correct thing to do. (also mentioned here: https://github.com/SBC-Utrecht/pytom-match-pick/discussions/266#discussioncomment-13735010)

This does **not** deal with the defocus handedness, as this is set by the `AreAnglesInverted` flag of the xml, which we currently not read. It only deals with the tomogram reconstruction handedness.

This would be particularly bad if the sampling wasn't symmetric with respect to angles.

[EDIT: we ended up testing this, see [this comment](https://github.com/SBC-Utrecht/pytom-match-pick/pull/334#issuecomment-3958455104) text below is to keep history intact]

We haven't truly tested this but reasoned our way in the following way:
1) we know our angles are correct for an AreTomo reconstructed tomogram with their tilt-angles
2) we know that with the same tilt-angles Warp gives us a reconstruction that is mirrored
2b) we know that flipping the defocus handedness does not change anything other than the  `AreAnglesInverted` flag. and leads to tomograms that are always mirrored compared to AreTomo
3) if we multiply the angles with -1 in the xml, we do end up with a reconstruction that has the same handedness as AreTomo
4) Hence, for our internal angles to match the tomogram from Warp, we should multiply the tilt-angles with -1

@shahpnmlab do you agree with this or have other strong opinions? For our info, did you ever test it with a non-symmetric sampling, as on symmetric sampling this would barely show up?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.12.2"
+version = "0.13.0"
 description = "PyTOM's GPU template matching module as an independent package"
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/src/pytom_tm/io.py
+++ b/src/pytom_tm/io.py
@@ -713,7 +713,8 @@ def parse_warp_xml_data(
     def _flatten(t):
         return [float(item) for sublist in t for item in sublist if item.strip()]
 
-    flattened_tilt_angles = _flatten(tilt_angles)
+    # -1 multiplication is needed to swap between warp internal convention to pytom
+    flattened_tilt_angles = [-i for i in _flatten(tilt_angles)]
     flattened_tilt_dose = _flatten(tilt_dose)
 
     ctf_data = [
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -5,6 +5,7 @@ import contextlib
 from tempfile import TemporaryDirectory
 import numpy as np
 import mrcfile
+from lxml import etree
 
 from pytom_tm.dataclass import CtfData, RelionTiltSeriesMetaData
 from pytom_tm.io import (
@@ -69,6 +70,20 @@ class TestWarpXMLParser(unittest.TestCase):
         voxel_size, ts_metadata = parse_warp_xml_data(WARP_XML, TEST_TOMOGRAM)
         for ctf in ts_metadata.ctf_data:
             self.assertTrue(10e-6 >= ctf.defocus >= 0.1e-6)
+
+    def test_correct_angle_sign(self):
+        voxel_size, ts_metadata = parse_warp_xml_data(WARP_XML, TEST_TOMOGRAM)
+        # grab raw xml data
+        tree = etree.parse(WARP_XML)
+        tilt_angle_nodes = tree.findall(".//Angles")
+        angles = [
+            float(j)
+            for i in tilt_angle_nodes
+            for j in i.text.split("\n")
+            if i.text.strip()
+        ]
+        for a, b in zip(ts_metadata, angles):
+            self.assertEqual(a, -b)
 
 
 class TestBrokenMRC(unittest.TestCase):
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -82,7 +82,7 @@ class TestWarpXMLParser(unittest.TestCase):
             for j in i.text.split("\n")
             if i.text.strip()
         ]
-        for a, b in zip(ts_metadata, angles):
+        for a, b in zip(ts_metadata.tilt_angles, angles):
             self.assertEqual(a, -b)
 
 


McHaillet:
Nice! Thanks for the PR and very clear explanation.
Also just mentioning that according to @builab this improves the performance.

McHaillet (approved):
Looks good to me review-wise and good that you updated the version. But maybe let's wait a bit until we hear something from others.

shahpnmlab:
Hi Sander, I dont have any specific complaints. Happy to test the current changes to see if it yields sharper peaks.

sroet:
Hey @shahpnmlab 

> Happy to test the current changes to see if it yields sharper peaks.

That would be great! could you try it out and report back? (specifically with an asymmetric wedge if that's possible as we are pretty sure the effect is more pronounced when the wedge isn't symmetric)

shahpnmlab:
Will do. Currently bogged down with multiple deadlines, so will not be any time soon unfortunately...

shahpnmlab:
Dear Sander, I tried multiplying the TiltAngles by -1 and rerunning TM, and atleast in my hands there is a weakening of the CC-peaks.
image
 
 The weakening of the CC-peaks in my case could be due to the inversion of the geometric handedness viz. my template (or a defocus inversion, I forget which way around it goes). I could re-run the experiment by inverting the chirality of my template, but I reckon we'd be back where we started. Eitherway, I am not sure if the proposed update significantly changes the workflow. One would still have to run a test TM experiment to determine for their setup what the appropriate transforms should be. Hope this helps!

sroet:
@shahpnmlab, thanks for the feedback, seems like we need to run our own testing as to why this is happening as it doesn't make sense to me.

sroet:
Alright, this is going to be a long write up, TLDR: this PR is correct and should be merged.

TLDR image (from @McHaillet):


## setup
To test this properly I decided to do template matching on a tomogram with only negative tilts and the 0 tilt (according to the mdocs). I have done everything in warptools, effectively following [the tutorial](https://warpem.github.io/user_guide/warptools/quick_start_warptools_tilt_series/) but using a tomogram of ribosomes instead.

## Tomogram reconstruction
I only gave warptools access to the `.tif` files of the negative tilts (and 0) and a modified `.mdoc` which was stripped of any data corresponding to the possitive tilts.
Other things that were different from the tutorial:
- pixel size, exposure, etc were different for the collected tilt series
- I did let warptools invert the angles on import of the frame series, and did not set a min_intensity
- I did not set a `--range_high`  for the `ts_ctf` command
- I gave the `ts_reconstruct` the `--dont_invert` flag as I knew the ribosomes were also dark in my tilts and the template was also dark.
- I stopped after `ts_reconstruct` created a tomogram

The reconstruction is "not great" (I guess there's a reason people don't tilt only one way for tomography 😅) but that doesn't matter for the actual TM results
slice 74
slice 149
slice 247

## TM results
I ran the TM with:
- both the main branch and this PR
- both with a regular and z-flipped template 
- with a spherical mask 
- grabbed everything directly from the warp xml
- gave the following extra options: ` --particle-diameter 150 --per-tilt-weighting --warp-xml-file ../warp_processing/Sample1_TS048.xml -g 0 1 2 3 --voxel-size-angstrom 10 -s 2 2 1`

When I then looked at the result, it was weird. The score volume of the main branch **did not change with the Z-height**
(same frames as above, top is main branch, bottom is this PR, top-left and bottom-right are z-flipped templates)

slice 74
slice 149
slice 247

## Why does this happen?
I ended up having a discussion with @McHaillet which drew the following amazing picture (recon is reconstruction):


As I did have the 0 tilt in there, using the wrong angle sign would only overlap on that tilt (and be 0 on all the other tilts, due to multiplication with 0). 
The 0 tilt lays flat in the xy plane and if we map it back from fourier space to real space, you end up just that image duplicated into the Z direction (as if you did back projection with just 1 image). This happens for the main branch and therefor reads in the wrong angles compared to the reconstructed tomogram, which is then fixed in this PR

McHaillet (approved):
Nice write-up! Agreed from my side!

shahpnmlab:
Thanks for the deepdive @sroet. I agree with the edits, as mentioned, it does not operationally change the outcomes for the user as one needs to check the process on at least one tiltseries with different handed templates to see which gives higher CC peaks. 
On a different note I had not realised that one needed to explicity call `--per-tilt-weighting`, I am going to give that a go and see if that improves the shape of the CC-peaks. (I have been omitting this flag, which results in the peaks being stretched along the Z-dim and I wonder if the inclusion of this flag will tighten up the shape of the peak).

McHaillet:
> On a different note I had not realised that one needed to explicity call `--per-tilt-weighting`, I am going to give that a go and see if that improves the shape of the CC-peaks. (I have been omitting this flag, which results in the peaks being stretched along the Z-dim and I wonder if the inclusion of this flag will tighten up the shape of the peak).

I think this flag is overwritten to True when a Warp xml is provided! If not, we should, because with the Warp metadata we have all the info to construct the right 3D-CTF.

McHaillet:
It is indeed: https://github.com/SBC-Utrecht/pytom-match-pick/blob/6d49915cd71c0dd6d40c6f09e977e278cc127150/src/pytom_tm/dataclass.py#L107-L110

McHaillet:
So the flag will actually be overwritten, we should probably update the CLI help

sroet:
> So the flag will actually be overwritten, we should probably update the CLI help

yeah it is silently dropped (like a bunch of other input) when either giving a relion5 starfile or a warp input

We should probably make an issue for that to at least raise a warning if the setting is identical as user input (as in my case) and otherwise error out or something

McHaillet:
yep sounds good

PR #336, opened 2026-02-27
reimplement job splitting to allow for arbitrary number of gpus

sroet:
supersedes #330 

Pytom can now handle almost any combination of volume splits numbers and gpu numbers.
It is done in the following priority:
1. every gpu gets the same number of tasks
2. we try to load (partial) tomograms the least amount of times

The one edge case we can't resolve is if the number of gpus is bigger than the number of volume splits times the number of sampled angles as that is the smallest quantity we have for subjobs. It errors out if that happens

### DIFF
--- a/src/pytom_tm/parallel.py
+++ b/src/pytom_tm/parallel.py
@@ -4,6 +4,7 @@ import logging
 import queue
 import time
 import contextlib
+from math import lcm
 from multiprocessing.managers import BaseProxy
 from functools import reduce
 from pytom_tm.tmjob import TMJob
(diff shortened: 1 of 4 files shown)


tests/test00_parallel.py
```diff
@@ -114,3 +114,56 @@ def test_parallel_manager(self):
...
+        # test more volume splits than gpus
+        # divisable
+        job = self.job.copy()
+        jobs = split_job_efficiently(job, (2, 2, 2), 4)
+        # jobs should be the same as total volume splits
+        self.assertEqual(len(jobs), 8)
+
+        # not divisible
+        job = self.job.copy()
+        jobs = split_job_efficiently(job, (3, 3, 1), 4)
```
McHaillet: What happens here? The comment is not correct here.

Number of pieces should be 9 instead of 6. How does the splitting work in that case if we go from 9 to 12? Are only a few of the subvolumes further split by angle?

src/pytom_tm/parallel.py
```diff
@@ -55,6 +56,50 @@ def gpu_runner(
                 break
 
 
+def split_job_efficiently(
```
McHaillet, line 59: Really clean rewrite!

tests/test00_parallel.py
```diff
@@ -114,3 +114,56 @@ def test_parallel_manager(self):
...
+
+        # test angle split only
+        job = self.job.copy()
+        jobs = split_job_efficiently(job, (1, 1, 1), 6)
+        self.assertEqual(len(jobs), 6)
+
+        # test volume split only
+        job = self.job.copy()
+        jobs = split_job_efficiently(job, (2, 2, 1), 4)
+        self.assertEqual(len(jobs), 6)
```
McHaillet:
```suggestion
        self.assertEqual(len(jobs), 4)
```

Should be 4 instead of 6 with a volume split of (2,2,1)

McHaillet (commented):
Just noticed a few tests that seemed off to me.

McHaillet:
Sorry, I just saw this was not review ready yet

McHaillet:
Maybe ignore my comments, if you were still updating it

sroet:
no worries, it should be done now (caught most of the things you mentioned in testing indeed)

tests/test00_parallel.py
sroet: You were, right, somehow my brain made volume splits additive (summing to 6, ignoring the 1) instead of multiplicative :facepalm:

McHaillet:
I really like the rewrite. Just find the lcm and increase angular splits until we find something that distributes nicely. Especially since the angular search split should have almost no latency.

src/pytom_tm/parallel.py
sroet: Thanks, was very happy when I remembered the `lcm` function :smile:

McHaillet: this is where your AoC experience comes in handy!

McHaillet (approved):
LGTM I can easily follow the logic and the tests cover the important cases!

sroet:
> Not entirely sure what you mean by "almost no latency". I tried to keep the volume loading to the minimum while keeping the same number of jobs and job size per gpu

[response of McHaillet]:
> I meant: splitting the angular search of single volume over N gpu's should have a roughly linear scaling in runtime. While with volume splitting, the subvolumes always require overhang to get similar results which expand the computation.

sroet:
> One question that I have, should we log somewhere the numbers of jobs we will be running?

Will leave this for another PR (we should probably look at a more readable logging for MPI than tqdm anyway)
Merging now

PR #337, opened 2026-03-02
change help formatter to ArgumentDefaultsHelpFormatter

### LINKED ISSUE #321: add defaults to usage page
Noticed that the `Usage` page on the website does not list defaults for arguments. It would be handy to show it.

sroet:
should close #321 

- [x] change help printer to include defaults
- [x] change doc building to not skip over defaults anymore 
- [x] test out new help messages
- [x] change logging defaults to the string "INFO" instead of 20 (allowed as logging level input) 
- [x] change random seed generation to outside of defaults to prevent people from thinking the one in the help message is constant and not regenerated every time you run the code 
- [x] added smoke tests for the entry points of `estimate_roc` and `extract_candidates` to make sure log defaults are sane (resulting in 100% coverage of entry_points.py)

updated usage page: 
[index.html](https://github.com/user-attachments/files/25687772/index.html) (click "skip to content" to see the new help messages, you can also locally build the whole site if you do want the icons etc to work)

### DIFF
--- a/tests/test_entry_points.py
+++ b/tests/test_entry_points.py
@@ -79,9 +79,13 @@ class TestEntryPoints(unittest.TestCase):
     @classmethod
     def setUpClass(cls) -> None:
         TEST_DATA.mkdir(parents=True)
-        io.write_mrc(TEMPLATE, np.zeros((5, 5, 5), dtype=np.float32), 1)
-        io.write_mrc(MASK, np.zeros((5, 5, 5), dtype=np.float32), 1)
-        io.write_mrc(TOMOGRAM, np.zeros((10, 10, 10), dtype=np.float32), 1)
+        volume = np.zeros((10, 10, 10), dtype=np.float32)
+        template = np.ones((5, 5, 5), dtype=np.float32)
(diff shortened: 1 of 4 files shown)


tests/test_entry_points.py
```diff
@@ -409,3 +415,71 @@ def start(arg_dict):  # simplify run
...
+            "--amplitude-contrast": "0.08",
+            "--spherical-aberration": "2.7",
+            "--voltage": "300",
+            "--tomogram-ctf-model": "phase-flip",
+            "-g": "0",
+        }
+        # generate match
+        entry_points.match_template(prep_argv(match_defaults))
+        tomo_id = f"{TOMOGRAM.stem}"
+        print(f"{str(self.outputdir / f'{tomo_id}_job.json')}")
```
McHaillet: Is this print supposed to be here? Or a remnant from debugging?

McHaillet (approved):
LGTM, just noticed one print statement that felt out of place

tests/test_entry_points.py
sroet: thanks for catching that, definitely debugging

PR #339, opened 2026-03-03
use safer init values for initial volumes

### LINKED ISSUE #338: Make init values safer for angle and score volumes
I ran into an issue with the tests I was writing for #337 that the angle index was set to -1000.

This was only an error for that test as the angle list is tiny.
However for a real system this might lead to an actual angle being selected that wasn't the highest at all.

We either have to filter out negative angle indices or chose different initialization values so we actually error out if we see them.

sroet:
as -1000 is a valid index in a lot of angle lists, this might lead to weird results.
~Try setting to NaN instead~

- [ ] ~change init value to NaN from -1000~
- [ ] ~make sure the NaNs are not propagated in statistics~

[Edit]: NaNs were propagating all over the place, decided to swap with `-inf` instead
- [x] change init value from -1000 to -inf both in the main TMJob and when merging subjobs
- [x] make sure the regular and tophat extraction can deal with -inf in the data
- [x] specifically for the tophat transform, we set the value to 0 (background) and then ignore them in the histogram 

closes #338

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -1,4 +1,5 @@
 from packaging import version
+import math
 import pandas as pd
 import numpy as np
 import numpy.typing as npt
@@ -215,6 +216,17 @@ def extract_particles(
         symmetry=job.rotational_symmetry,
     )
 
+    # Check for invalid values or NaNs
+    for x in ["variance", "std"]:
+        if math.isinf(job.job_stats[x]) or math.isnan(job.job_stats[x]):
+            raise ValueError(
+                f"job stat '{x}' is NaN or inf, please check your volumes and update "
+                "the job_stats json accordingly"
+            )
+
+    sigma = job.job_stats["std"]
+    search_space = job.job_stats["search_space"]
+
     if tophat_filter:  # constrain the extraction with a tophat filter
         predicted_peaks = predict_tophat_mask(
             score_volume,
@@ -285,8 +297,6 @@ def extract_particles(
     score_volume[:, -particle_radius_px:, :] = 0
     score_volume[:, :, -particle_radius_px:] = 0
 
-    sigma = job.job_stats["std"]
-    search_space = job.job_stats["search_space"]
     if cut_off is None:
         # formula Rickgauer et al. (2017, eLife):
         # N**(-1) = erfc( theta / ( sigma * sqrt(2) ) ) / 2
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -327,11 +327,11 @@ def extract_particles(
     scores = []
 
     for _ in tqdm(range(n_particles)):
-        ind = np.unravel_index(score_volume.argmax(), score_volume.shape)
+        ind = np.unravel_index(np.nanargmax(score_volume), score_volume.shape)
 
         lcc_max = score_volume[ind]
 
-        if lcc_max <= cut_off:
+        if lcc_max <= cut_off or np.isnan(lcc_max):
             break
 
         scores.append(lcc_max)
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -79,7 +79,7 @@ def predict_tophat_mask(
     # to exclude them from the histogram
     if np.any(np.isneginf(score_volume)):
         finite_indices = np.where(np.isfinite(score_volume))
-        score_volume = np.nan_to_num(neginf=0.0)
+        score_volume = np.nan_to_num(score_volume, neginf=0.0)
     else:
         finite_indices = None, None, None
 
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -75,13 +75,21 @@ def predict_tophat_mask(
     if score_volume.dtype == np.float16:
         score_volume = score_volume.astype(np.float32)
 
+    # convert -inf (masked) scores to background 0 and kep track of them
+    # to exclude them from the histogram
+    if np.any(np.isneginf(score_volume)):
+        finite_indices = np.where(np.isfinite(score_volume))
+        score_volume = np.nan_to_num(neginf=0.0)
+    else:
+        finite_indices = None, None, None
+
     tophat = ndimage.white_tophat(
         score_volume,
         structure=ndimage.generate_binary_structure(
             rank=3, connectivity=tophat_connectivity
         ),
     )
-    y, bins = np.histogram(tophat.flatten(), bins=bins)
+    y, bins = np.histogram(tophat[finite_indices].flatten(), bins=bins)
     bin_centers = (bins[:-1] + bins[1:]) / 2
     x_raw, y_raw = (
         bin_centers[2:],
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -76,8 +76,8 @@ class TemplateMatchingPlan:
 
         # Initialize result volumes
         self.ccc_map = cp.zeros(volume_shape, dtype=cp.float32)
-        self.scores = cp.ones(volume_shape, dtype=cp.float32) * -1000
-        self.angles = cp.ones(volume_shape, dtype=cp.float32) * -1000
+        self.scores = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
+        self.angles = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
 
         self.random_phase_template_texture = None
         self.noise_scores = None
@@ -87,7 +87,7 @@ class TemplateMatchingPlan:
                 interpolation="filt_bspline",
                 device=f"gpu:{device_id}",
             )
-            self.noise_scores = cp.ones(volume_shape, dtype=cp.float32) * -1000
+            self.noise_scores = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
 
         # wait for stream to complete the work
         cp.cuda.stream.get_current_stream().synchronize()
--- a/src/pytom_tm/matching.py
+++ b/src/pytom_tm/matching.py
@@ -76,8 +76,8 @@ class TemplateMatchingPlan:
 
         # Initialize result volumes
         self.ccc_map = cp.zeros(volume_shape, dtype=cp.float32)
-        self.scores = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
-        self.angles = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
+        self.scores = cp.ones(volume_shape, dtype=cp.float32) * float("-inf")
+        self.angles = cp.ones(volume_shape, dtype=cp.float32) * float("-inf")
 
         self.random_phase_template_texture = None
         self.noise_scores = None
@@ -87,7 +87,7 @@ class TemplateMatchingPlan:
                 interpolation="filt_bspline",
                 device=f"gpu:{device_id}",
             )
-            self.noise_scores = cp.ones(volume_shape, dtype=cp.float32) * float("nan")
+            self.noise_scores = cp.ones(volume_shape, dtype=cp.float32) * float("-inf")
 
         # wait for stream to complete the work
         cp.cuda.stream.get_current_stream().synchronize()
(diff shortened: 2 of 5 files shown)


McHaillet:
Good to see all the tests still pass with this update. I am still a bit anxious about this update: are there any code paths that might break because of this? What if there are -inf value in the score map and the tophat transform is done on it for example? 

Maybe my anxiety is unwarranted, because cross correlation should always provide values between -1 and 1, meaning that the update kernel (that writes from ccc_map to scores) should at least insert values larger than -inf for each voxel.  Which means the original -inf will always be overwritten even when searching a single rotation.

sroet:
> Maybe my anxiety is unwarranted, because cross correlation should always provide values between -1 and 1, meaning that the update kernel (that writes from ccc_map to scores) should at least insert values larger than -inf for each voxel. Which means the original -inf will always be overwritten even when searching a single rotation.

It should always be overwritten, but:
1. the MRC file writer will give a warning if you write an MRC file with +-inf values in them (visible in the entry point test logs)
2. It would now break loudly instead of silently giving wrong results:
- for the angles you now will get a `OverflowError: cannot convert float infinity to integer` if we try to load an angle index that has not been overwritten, instead of the possible valid `-1000`
- for the scores: the standard deviation will become `NaN`, (and the mean `-inf`) which should break most/all of our extraction code paths loudly

McHaillet:
Thanks for addressing the concerns. That's a great point: better to break loudly instead of silently for something that we anyway assumed to be overwritten in the current code.

McHaillet (approved):
Clean fix and all the tests pass, so LGTM

sroet:
Extract might not actually fail loudly if those stats are NaN/-inf looking at #331, will  add an explicit check in the extraction job

sroet:
Alright, some other weird behavior we need to have a discussion about: what should the scores be when stuff is filtered by the tomogram mask? The `merge_subjobs` code seems inconsistent (setting to `-1` for angle split jobs, but setting to `0` for volume split jobs)
https://github.com/SBC-Utrecht/pytom-match-pick/blob/b22d5ac4602644f7301267d914f83602f91fffd0/src/pytom_tm/tmjob.py#L837-L852

sroet:
@McHaillet Thanks for bringing up your concerns, please have another look! I would also like your opinion on [this comment](https://github.com/SBC-Utrecht/pytom-match-pick/pull/339#issuecomment-3995905928) if it is not too much trouble

McHaillet:
Was it clear what caused the NaN's in the end in #331? Was it because of the mask?

McHaillet:
> @McHaillet Thanks for bringing up your concerns, please have another look! I would also like your opinion on [this comment](https://github.com/SBC-Utrecht/pytom-match-pick/pull/339#issuecomment-3995905928) if it is not too much trouble

Yea that is very inconsistent. To me 0 make the most sense as that would indicate no correlation. -1 would be inverse correlation which is weird. For a masked area we have no idea, so then 0 seems most logical

sroet:
> Was it clear what caused the NaN's in the end in #331? Was it because of the mask?

probably, most of my tests would end up with `inf` instead of `NaN`, the only option if found is that this line the code:
https://github.com/SBC-Utrecht/pytom-match-pick/blob/b22d5ac4602644f7301267d914f83602f91fffd0/src/pytom_tm/matching.py#L303-L305

 has 0s for both `self.plan.ccc_map * roi_mask` and `roi_size`, which is most likely due to mask shenanigans

image

McHaillet:
Strange... The roi_size should never be zero I think

McHaillet:
The roi_mask name is just the unique area of the subvolume after tiling the tomogram.

src/pytom_tm/extract.py
```diff
@@ -215,6 +216,17 @@ def extract_particles(
         symmetry=job.rotational_symmetry,
     )
 
+    # Check for invalid values or NaNs
```
McHaillet, line 227: This makes sense. Should the scores volume also be checked for Nans/infs? Or did we say that shouldn't happen and we let it error?

sroet: If the score volumes have NaNs/infs these values should also be set to NaN/inf

sroet:
initializing to 0 for the scores in the angle split job does not work as it would override any real negative correlation values (as we are looking for the best score), so set it to `-inf` instead

src/pytom_tm/extract.py
McHaillet: right!

McHaillet:
> initializing to 0 for the scores in the angle split job does not work as it would override any real negative correlation values (as we are looking for the best score), so set it to `-inf` instead

Ah true! All the split jobs should cover the whole volume, so no `-inf` should be left after merging. I guess this is also a case of better to error out hard if our assumptions are not working. 

I guess that only leaves the case of a subvolume where the tomogram mask (segmentation) has 0's. Because that subjob is skipped and, judging by the current code, means the -inf's will be left there. `-inf` should never be annotated during extraction though as it's values are smaller than any float.

sroet:
> I guess that only leaves the case of a subvolume where the tomogram mask (segmentation) has 0's. Because that subjob is skipped and, judging by the current code, means the -inf's will be left there. -inf should never be annotated during extraction though as it's values are smaller than any float.

Yeah, it should never be extracted and as those sub jobs are not run at all, they also don't count against the statistics, so no possible `inf` or `NaN` in those under normal circumstances

McHaillet:
It means that split jobs, with a tomogram mask that results in empty subjobs, will have results with nan's saved in the score map. Two things:
1. can mrcfile save -inf values?
2.  the tophat transform runs directly on the score map and apparently only raises a warning (see below). But not sure what happens with the cutoff estimation if there are nans.

Wanted to test what happens with inf's, because the tophat transform runs directly on the score map that can contain inf. 

```
a = np.zeros((100,100))
a[:] = float('-inf')
a[50,50] = 1
x = ndimage.white_tophat(a, structure=ndimage.generate_binary_structure(rank=2, connectivity=1))
```

It works, but gives this warning:

```
/home/marten/miniconda3/envs/plot/lib/python3.12/site-packages/scipy/ndimage/_morphology.py:1898: RuntimeWarning: invalid value encountered in subtract
  np.subtract(input, tmp, out=tmp)
```

Everything is nan, except for the peak at (50,50).

sroet:
> can mrcfile save -inf values?

Yes, it will also warn with a pretty clear message:
```
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mrcfile
>>> import numpy as np
>>> a = np.array([[0, 2., 4., float('-inf')]]*4, dtype=np.float32)
>>> with mrcfile.new('test.mrc') as mrc:
...   mrc.set_data(a)
...
[redacted]/miniconda3/envs/pytom_tm/lib/python3.12/site-packages/mrcfile/mrcobject.py:606: RuntimeWarning: Data array contains infinite values
  warnings.warn("Data array contains infinite values", RuntimeWarning)
[redacted]/miniconda3/envs/pytom_tm/lib/python3.12/site-packages/numpy/_core/_methods.py:191: RuntimeWarning: invalid value encountered in subtract
  x = asanyarray(arr - arrmean)
```

> Everything is nan, except for the peak at (50,50).

NaNs are bigger or than normal scores (looking at a test it seems they pop up in argmax, will see if I can implement a catch for that)

```
>>> b
array([[ nan,  inf,  inf, -inf],
       [ nan,  inf,  inf, -inf],
       [ nan,  inf,   1., -inf],
       [ nan,  inf,  inf, -inf]], dtype=float32)
>>> np.unravel_index(b.argmax(), b.shape)
(np.int64(0), np.int64(0))
```

will swap to nanargmax instead and deal with the warning on [the reference page](https://numpy.org/doc/stable/reference/generated/numpy.nanargmax.html) as it might still return an index of a NaN if it is just NaN and -inf

sroet:
> It means that split jobs, with a tomogram mask that results in empty subjobs, will have results with nan's saved in the score 

That shouldn't be the case as the merge jobs part should reset that to 0 (those jobs don't exist, we never even create subjobs for them so the default value of 0 is never overwritten):
https://github.com/sroet/pytom-template-matching-gpu/blob/1f54f1c8d36b05b0ee4ac1e803d2d3e55ce7a097/src/pytom_tm/tmjob.py#L847-L874

McHaillet:
> > It means that split jobs, with a tomogram mask that results in empty subjobs, will have results with nan's saved in the score
> 
> That shouldn't be the case as the merge jobs part should reset that to 0 (those jobs don't exist, we never even create subjobs for them so the default value of 0 is never overwritten): https://github.com/sroet/pytom-template-matching-gpu/blob/1f54f1c8d36b05b0ee4ac1e803d2d3e55ce7a097/src/pytom_tm/tmjob.py#L847-L874

But you changed it -inf now, right?

McHaillet:
> will swap to nanargmax instead and deal with the warning on [the reference page](https://numpy.org/doc/stable/reference/generated/numpy.nanargmax.html) as it might still return an index of a NaN if it is just NaN and -inf

good idea

sroet:
> > > It means that split jobs, with a tomogram mask that results in empty subjobs, will have results with nan's saved in the score
> > 
> > That shouldn't be the case as the merge jobs part should reset that to 0 (those jobs don't exist, we never even create subjobs for them so the default value of 0 is never overwritten): https://github.com/sroet/pytom-template-matching-gpu/blob/1f54f1c8d36b05b0ee4ac1e803d2d3e55ce7a097/src/pytom_tm/tmjob.py#L847-L874
> 
> But you changed it -inf now, right?

The merge default is still 0, will change it to -inf to make it clear we have no info on that

McHaillet:
Just to summarize to see if I understand correctly.

* in the case of split jobs, the scores and angles arrays are initialized with -inf
* when a cell segmentation is provided some subvolumes will just be skipped, this means -inf will be left in the array after merging
* this wil raise a warning with mrcfile (makes sense) but maybe we should suprress it because we actually should expect this to happen
* normal extraction should work fine as the statistics are okay and -inf is smaller than actual scores
* tophat will probably fail, the tophat transform should be okay, but the cutoff calculation likely fails due to nan's after the transform

sroet:
> Just to summarize to see if I understand correctly.
> 
> * in the case of split jobs, the scores and angles arrays are initialized with -inf
> * when a cell segmentation is provided some subvolumes will just be skipped, this means -inf will be left in the array after merging
> * this wil raise a warning with mrcfile (makes sense) but maybe we should suprress it because we actually should expect this to happen
> * normal extraction should work fine as the statistics are okay and -inf is smaller than actual scores
> * tophat will probably fail, the tophat transform should be okay, but the cutoff calculation likely fails due to nan's after the transform

Not the current state (merging volume split scores are still initialized to 0 at the moment), but will make sure that is changed to -inf when I am in the office

McHaillet:
> Not the current state (merging volume split scores are still initialized to 0 at the moment), but will make sure that is changed to -inf when I am in the office

I know, but this is what happens after this update!

Maybe  `predict_tophat_mask()` can just replace Nan's with zero after doing the transform. Those locations should be background.

McHaillet:
In matching.py both scores/angles are initialized with -inf, but in the merging the scores with -inf and the angles with inf

sroet:
> In matching.py both scores/angles are initialized with -inf, but in the merging the scores with -inf and the angles with inf

Yeah, that had to do with the fact that the following line misbehaved when we were still initializing the scores with 0, will try initializing with -inf instead and see if anything breaks

https://github.com/SBC-Utrecht/pytom-match-pick/blob/b22d5ac4602644f7301267d914f83602f91fffd0/src/pytom_tm/tmjob.py#L845

sroet:
> Maybe predict_tophat_mask() can just replace Nan's with zero after doing the transform. Those locations should be background.

Do you want me to include that as part of this PR or for a next one? I don't mind either and have a pretty clear idea on how to deal with the tophat transform (setting to 0 for the tophat and then excluding from the histogram))

McHaillet:
Let's include it in this one. Then this PR is a full fix of the issue

sroet:
All set, @McHaillet please have another look to see if you agree with this fix!

McHaillet (approved):
Nice, looks good and all edges cases covered for the extraction!

sroet:
Thank you for the in depth discussion and catching the edge cases!

PR #341, opened 2026-03-11
Deal with dropped args

### LINKED ISSUE #335: Make sure we mention to the user when explicit user input is overwritten/dropped
> So the flag will actually be overwritten, we should probably update the CLI help

yeah it is silently dropped (like a bunch of other input) when either giving a relion5 starfile or a warp input

We should probably make an issue for that to at least raise a warning if the setting is identical as user input (as in my case) and otherwise error out or something

_Originally posted by @sroet in https://github.com/SBC-Utrecht/pytom-match-pick/issues/334#issuecomment-3960257996_

sroet:
closes #335 
for now I decided to output a warning
- [x] implement code
- [x] implement test
- [x] test long/short/abbreviated options

When reviewing, please give me your opinion on the following alternatives:
- Error when there are dropped args at all
- Error normally, but let users provide a flag if they want to use it to override warp/relion data instead
- Warn normally, but let users provide a flag if they want to use it to override warp/relion data

For the last 2 options do you want to first merge this PR or already implement the new flag in this PR as well (would require a big rewrite as the whole ctf data block that happens before the relion/warp parsing then needs to account for missing args) https://github.com/SBC-Utrecht/pytom-match-pick/blob/9fd5388fb56d8122cd26c507114e88741e018f3e/src/pytom_tm/entry_points.py#L1014-L1035

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -1033,7 +1033,7 @@ def match_template(argv=None):
             )
             for defocus in args.defocus
         ]
-
+    dropped_args = []
     if args.relion5_tomograms_star is not None:
         voxel_size, ts_metadata = parse_relion5_star_data(
             args.relion5_tomograms_star,
@@ -1041,6 +1041,18 @@ def match_template(argv=None):
             phase_flip_correction=phase_flip_correction,
             phase_shift=args.phase_shift,
         )
+        dropped_args += [
+            "--defocus",
+            "--amplitude_contrast",
+            "--voltage",
+            "--spherical-aberration",
+            ("-a", "--tilt-angles"),
+            "--per-tilt-weighting",
+            "--warp-xml-file",
+            "--voxel-size-angstrom",
+            "--dose-accumulation",
+            "--defocus-handedness",
+        ]
 
     elif args.warp_xml_file is not None:
         voxel_size, ts_metadata = parse_warp_xml_data(
@@ -1050,6 +1062,16 @@ def match_template(argv=None):
         )
         # Replace is needed here to rerun the sanity checking
         ts_metadata = ts_metadata.replace(defocus_handedness=args.defocus_handedness)
+        dropped_args += [
+            "--defocus",
+            "--amplitude_contrast",
+            "--voltage",
+            "--spherical-aberration",
+            ("-a", "--tilt-angles"),
+            "--per-tilt-weighting",
+            "--voxel-size-angstrom",
+            "--dose-accumulation",
+        ]
 
     else:
         if tilt_angles is None:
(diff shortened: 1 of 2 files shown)


src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
             per_tilt_weighting=args.per_tilt_weighting,
         )
 
+    # Warn for dropped args, use new parser to deal with possible shorthand
```
McHaillet, line 1098: Line 1098 to 1115 are not so easy to interpret.

* the names test_input and test_args are confusing because my first intuition is that they are stored for unittests
* the parsing is neccessary to check for shorthand commands, correct?

McHaillet:
> * Error when there are dropped args at all
> 
> * Error normally, but let users provide a flag if they want to use it to override warp/relion data instead
> 
> * Warn normally, but let users provide a flag if they want to use it to override warp/relion data

Tough one... My first thought is to just error out completely. If the metadata is there, it seems best to trust that all the options can be set correctly. And it prevents over complicating the code.

src/pytom_tm/entry_points.py
sroet: > the names test_input and test_args are confusing because my first intuition is that they are stored for unittests

I understand, feel free to propose different names for this

> the parsing is neccessary to check for shorthand commands, correct?

yes indeed, as the shorthand is dependent on any of the other options (single letter shorthand can be allowed if there is no other long option that starts with the same letter)

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+    test_input = argparse.ArgumentParser(
```
McHaillet:
```suggestion
    check_input = argparse.ArgumentParser(
```

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+        add_help=False, argument_default=argparse.SUPPRESS
+    )
+    for arg in dropped_args:
+        if isinstance(arg, tuple):
+            short, long = arg
+            test_input.add_argument(
```
McHaillet:
```suggestion
            check_input.add_argument(
```

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+                short, long, action="store_const", const=f"{short}/{long}"
+            )
+        else:
+            test_input.add_argument(arg, action="store_const", const=f"{arg}")
```
McHaillet:
```suggestion
            check_input.add_argument(arg, action="store_const", const=f"{arg}")
```

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+    # drop any input values that were not options
+    test_argv = [i for i in argv if "-" in i]
```
McHaillet:
```suggestion
    check_argv = [i for i in argv if "-" in i]
```

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+    test_args, _ = test_input.parse_known_args(test_argv)
```
McHaillet:
```suggestion
    check_args, _ = check_input.parse_known_args(check_argv)
```

src/pytom_tm/entry_points.py
```diff
@@ -1067,6 +1095,24 @@ def match_template(argv=None):
...
+    for val in vars(test_args).values():
```
McHaillet:
```suggestion
    for val in vars(check_args).values():
```

sroet:
> My first thought is to just error out completely. 

Do you feel strongly enough to implement this now?

McHaillet:
Sorry for getting back to this late, I think we can merge this already. The warning is most important.

McHaillet (approved):

PR #342, opened 2026-03-16
deal with possible negative angles from healpix

sroet:
see: https://github.com/ntessore/healpix/issues/99

This mainly adds some checks and balances to make sure our angles are between 0 and 2π for Z1 and Z2, and between 0 and π for X

- [x] add (failing) test
- [x] fix code

### DIFF
--- a/tests/test_template_matching.py
+++ b/tests/test_template_matching.py
@@ -58,7 +58,7 @@ class TestTM(unittest.TestCase):
         )
         self.assertAlmostEqual(
             stats["std"],
-            0.005163,
+            0.005006,
             places=5,
             msg="Standard deviation of the search should be almost equal",
         )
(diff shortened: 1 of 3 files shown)


tests/test_template_matching.py
```diff
@@ -58,7 +58,7 @@ def test_search_spherical_mask(self):
         )
         self.assertAlmostEqual(
             stats["std"],
-            0.005163,
```
McHaillet, line 61: why did this change?

sroet: I don't know :sweat_smile: It might just be due to the change in order due to the "order of summation" inaccuracies in floating points... https://stackoverflow.com/questions/48957828/floating-point-arithmetic-why-would-order-of-addition-matter

another option would be that somewhere in our dependencies (I suspect voltools) there is an assumption about the ranges for these values

McHaillet: right, that seems reasonable, I guess its not a big deal. And your suggested change makes a lot of sense

McHaillet (approved):
LGTM

PR #343, opened 2026-03-16
update github actions

sroet:
The following warning started popping up in our tests:
```
Warning: Node.js 20 actions are deprecated. 
The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/upload-artifact@v4. 
Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. 
Please check if updated versions of these actions are available that support Node.js 24. 
To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file.
Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. 
For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

started by updating our standard actions to the newest available version

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -30,7 +30,7 @@ jobs:
       # Allow to check commit statusses
     steps:
       - name: Download Artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           name: coverage
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up Python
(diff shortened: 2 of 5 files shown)


McHaillet (approved):
Makes sense, LGTM

What about adding dependabot (https://github.com/teamtomo/torch-fourier-slice/blob/main/.github/dependabot.yml) to automatically keep them up to date?

sroet:
I am very worried about a supply chain attack like https://unit42.paloaltonetworks.com/github-actions-supply-chain-attack/ which we could be hit by by "auto-updating"

McHaillet:
From what I know, dependabot just makes PR's that propose to update the version of your actions. You still need to manually merge the PR. So, you can always check after it makes a proposal.
Or are you afraid of the bot having certain access to the repository?

sroet:
Exactly,  PRs from maintainers auto run on our hardware, as will any of the PRs that the bot proposes. As opposed to outside PRs that we will have to approve manually. To be fair that specific case would also hit us by overriding all the tags of the inflected action repo

McHaillet:
Okido!

PR #346, opened 2026-03-26
Move some imports to make GPU optional

MoritzWM:
Hello,
I'm using pytom-match-pick on a HPC cluster and love it. Though one issue I'm having is that I sometimes play with the false positive rate or cutoff and I don't want to submit that as GPU-bound jobs. The same applies to merging star files.

I figured I could just move around some imports to make it not fail when there's no GPU. It still requires Cuda for installation though.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pytom-match-pick"
-version = "0.14.0"
+version = "0.14.1"
 description = "PyTOM's GPU template matching module as an independent package"
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -895,6 +895,7 @@ class TMJob:
             with search statistics
         """
         from pytom_tm.matching import TemplateMatchingGPU
+
         tomo = read_mrc(self.tomogram)
         fast_tomo = np.zeros(
             tuple([next_fast_len(s, real=True) for s in tomo.shape]),
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -10,7 +10,6 @@ import json
 import logging
 from scipy.fft import next_fast_len, rfftn, irfftn
 from pytom_tm.angles import get_angle_list
-from pytom_tm.matching import TemplateMatchingGPU
 from pytom_tm.weights import (
     create_wedge,
     power_spectrum_profile,
@@ -895,6 +894,7 @@ class TMJob:
             angle map), when no volumes are returned the output consists of a dictionary
             with search statistics
         """
+        from pytom_tm.matching import TemplateMatchingGPU
         tomo = read_mrc(self.tomogram)
         fast_tomo = np.zeros(
             tuple([next_fast_len(s, real=True) for s in tomo.shape]),
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -17,7 +17,6 @@ from pytom_tm.angles import get_angle_list
 from pytom_tm.dataclass import CtfData, RelionTiltSeriesMetaData, TiltSeriesMetaData
 from pytom_tm.io import UnequalSpacingError, read_mrc, read_mrc_meta_data, write_mrc
 from pytom_tm.json import CustomJSONDecoder, CustomJSONEncoder
-
 from pytom_tm.weights import (
     create_gaussian_band_pass,
     create_wedge,
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -1,7 +1,6 @@
 import numpy as np
 import numpy.typing as npt
 import scipy.ndimage as ndimage
-import voltools as vt
 from pytom_tm.io import UnequalSpacingError
 from pytom_tm.dataclass import CtfData, TiltSeriesMetaData
 from itertools import pairwise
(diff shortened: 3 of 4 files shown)


sroet:
Hey @MoritzWM,

Thanks for the PR! I understand your use case, did you test that this is all it takes for your jobs to run?

Could you tell me:
1. Which of our entry points you normally try to run without a gpu? 
2. What is the error you get when you try to run on a non-gpu node with current main
3. What is the error you get when you try to run gpu code on a non-gpu node with your PR

First option is so I can think about what tests to add to make sure this doesn't break in the future
2 and 3 is just to make sure the error message doesn't get significantly more confusing than before.

sroet:
seems like our pre-commit checks are failing, do you mind:
1. running the steps under [developer install](https://github.com/SBC-Utrecht/pytom-match-pick?tab=readme-ov-file#developer-install), up to and including `pre-commit install`
2. run `ruff format` (after updating ruff in your conda install if that hasn't happened in a while)
3. commit and push the result

MoritzWM:
Hello @sroet,

sorry forgot to do the most basic things. Fixed the formatting.

Yes it seems that this is all that is required. I didn't test all options though, only a basic set. I will test a little more tomorrow to really make sure.

> 1. Which of our entry points you normally try to run without a gpu?

Only pytom_extract_candidates.py and pytom_merge_stars.py.

> 2. What is the error you get when you try to run on a non-gpu node with current main

```
pytom_create_template.py
  File "/gpfs/lsdf02/sd18g002/.helix/envs/pytom_dev/bin/pytom_create_template.py", line 6, in <module>
    sys.exit(pytom_create_template())
             ~~~~~~~~~~~~~~~~~~~~~^^
  File "/gpfs/lsdf02/sd18g002/Moritz/git/pytom-match-pick/src/pytom_tm/entry_points.py", line 148, in pytom_create_template
    from pytom_tm.template import generate_template_from_map
  File "/gpfs/lsdf02/sd18g002/Moritz/git/pytom-match-pick/src/pytom_tm/template.py", line 3, in <module>
    import voltools as vt
  File "/gpfs/lsdf02/sd18g002/.helix/envs/pytom_dev/lib/python3.14/site-packages/voltools/__init__.py", line 3, in <module>
    from .transforms import AVAILABLE_INTERPOLATIONS, AVAILABLE_DEVICES, scale, shear, rotate, translate, transform, affine
  File "/gpfs/lsdf02/sd18g002/.helix/envs/pytom_dev/lib/python3.14/site-packages/voltools/transforms.py", line 19, in <module>
    AVAILABLE_DEVICES = utils.get_available_devices()
  File "/gpfs/lsdf02/sd18g002/.helix/envs/pytom_dev/lib/python3.14/site-packages/voltools/utils/general.py", line 73, in get_available_devices
    for i in range(cupy.cuda.runtime.getDeviceCount()):
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "cupy_backends/cuda/api/runtime.pyx", line 420, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 423, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
```

> 3. What is the error you get when you try to run gpu code on a non-gpu node with your PR

Same error.

> First option is so I can think about what tests to add to make sure this doesn't break in the future 2 and 3 is just to make sure the error message doesn't get significantly more confusing than before.

I personally think the error message is fine. It says CUDA, it shows where the error is coming from and hiding the line where it is thrown will only make it harder to debug.

sroet:
Thanks for dealing with the feedback!

> Yes it seems that this is all that is required. I didn't test all options though, only a basic set. I will test a little more tomorrow to really make sure.

Please let me know if you still wanted to do this, or if I should go ahead and see if I can implement some tests for this?

> I personally think the error message is fine. It says CUDA, it shows where the error is coming from and hiding the line where it is thrown will only make it harder to debug.

Yeah, as long as it is failing hard inside the cuda backends it is fine, just wanted to make sure it doesn't lead to weird importerrors.

MoritzWM:
I tested on our HPC login node and it seems to work just fine. I just wanted to make sure I didn't make some stupid mistake.

One could maybe test that voltools and cupy are not loaded as modules when calling merge_stars and extract_particles?

sroet:
Hey @MoritzWM,

Thanks for checking! I am trying to design the tests. My idea is to mock out GPU availability somehow (probably setting `CUDA_VISIBLE_DEVICES` to an empty string) or by somehow mocking `cupy` (as both `pytom_tm` and `voltools` use that to interface with the gpus)

I am getting somewhere, but not quite there yet. Will push to your branch once I have figured out a test that fails on main, but works on your branch.

It is surprisingly hard to pretend to run on a machine without gpus even if the machine has them for one test and then have the GPU enabled for other tests. (Specifically struggling with python's import caching atm)😅.

sroet:
Alright, it took a while, but I've added a test I am happy with for this functionality. It fails on main and not on this branch and should be pretty robust.

>It is surprisingly hard to pretend to run on a machine without gpus even if the machine has them for one test and then have the GPU enabled for other tests. (Specifically struggling with python's import caching atm)😅.

The solution (that Claude Sonnet 5 came up with) is to sidestep the whole python import caching by calling the entry points from a sub-process, forcing a new (broken) import 

@McHaillet this is ready for a review by someone that is not me

McHaillet (approved):
LGTM, looks to me like a clean way to test it

sroet:
This is in, thanks for the fix (and your patience) @MoritzWM ! (Let us know if you need us to do a release with this fix or if it can wait until we completed some other stuff as well)

PR #349, opened 2026-04-09
Fix divide-by-zero in tilt_increment for duplicate tilt angles

### LINKED ISSUE #348: Divide-by-zero in tilt_increment with bidirectional tilt series (duplicate 0° angles)
## Bug

Template matching produces all-zeros score maps (-1000) when using a bidirectional tilt series that has two images at the same tilt angle (e.g., 0°).

## Cause

In `weights.py`, `_create_tilt_weighted_wedge()` computes:

```python
tilt_increment = min([abs(x - y) for x, y in pairwise(tilt_angles)])
```

Bidirectional schemes (0° → +50°, then 0° → -50°) acquire two images at 0°. When the angle list is sorted, these adjacent `0.0` entries produce `abs(0 - 0) = 0`, so `tilt_increment = 0`. Then:

```python
overlap_frequency = 1 / (tilt_increment * image_size)  # 1/0 = inf
```

This makes the ramp filter all zeros → wedge all zeros → template all zeros → NaN correlations → scores remain at -1000.

## To reproduce

Use any `.tlt` file with two identical angle entries (common in bidirectional tilt series) and enable `--per-tilt-weighting`.

## Suggested fix

Use unique sorted angles for computing `tilt_increment`:

```python
tilt_increment = min([abs(x - y) for x, y in pairwise(sorted(set(tilt_angles)))])
```

This computes the angular step from distinct tilt positions only. The per-tilt weighting loop that applies dose/CTF weights is unchanged and still uses all tilts including duplicates.

See #349.

felipemunoz8128:
Fixes #348.

Bidirectional tilt schemes acquire two images at 0° (one per half). When the tilt angle list is sorted, adjacent `0.0` entries produce a pairwise difference of 0, making `tilt_increment = 0` and `overlap_frequency = inf`. This zeros out the ramp filter, the wedge, and all correlation scores.

**Fix:** Use `sorted(set(tilt_angles))` to compute `tilt_increment` from unique angles only. One line changed. The per-tilt weighting loop is unchanged and still uses all tilts including duplicates for dose/CTF weighting.

### DIFF
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -548,7 +548,7 @@ def _create_tilt_weighted_wedge(
     tilt_weighted_wedge = np.zeros((image_size, image_size, image_size // 2 + 1))
 
     # create ramp weights to correct tilt summation for overlap
-    tilt_increment = min([abs(x - y) for x, y in pairwise(tilt_angles)])
+    tilt_increment = min([abs(x - y) for x, y in pairwise(sorted(set(tilt_angles)))])
     # Crowther freq. determines till what point adjacent tilts overlap in Fourier space
     overlap_frequency = 1 / (tilt_increment * image_size)
     freq_1d = (


sroet:
Hey @felipemunoz8128, thanks for the fix. It seems small enough and I can understand what issues would pop up.

I am a bit surprised people would be making two 0 degree images, as I don't think you would get any extra information for the extra dose on the sample.

Do you mind adding a test that makes sure this function does not return any NaN/inf when given a set of tilt angles with a duplicate number in it?  (can be 0, or a random pre-existing tilt, no real preference from me)

McHaillet:
> I am a bit surprised people would be making two 0 degree images, as I don't think you would get any extra information for the extra dose on the sample.

I am also confused by this. I thought one would start at 0 and go to -54, and then go from +3 to +54

PR #351, opened 2026-04-09
Fix inflated sigma when using random-phase noise correction

### LINKED ISSUE #350: Automated cutoff too high when using --random-phase-correction
## Bug

When `--random-phase-correction` is enabled, the automated extraction cutoff is too high, causing very few or zero particles to be extracted.

## Cause

In `matching.py`, `stats["variance"]` and `stats["std"]` are accumulated from the per-rotation `ccc_map` values during the angular search loop. When `--random-phase-correction` is enabled, the scores are then corrected:

```python
self.plan.scores = (self.plan.scores - self.plan.noise_scores) + self.plan.noise_scores.mean()
```

But sigma is computed from the pre-correction accumulated variance, not from the corrected scores. The noise correction reduces the effective noise in the score map, so the pre-correction sigma is inflated. The Rickgauer et al. cutoff formula then produces a threshold that is too high for the corrected score distribution.

## Suggested fix

After noise correction, recompute variance/std from the corrected score map within the ROI. The non-corrected path is unchanged.

See #351.

felipemunoz8128:
Fixes #350.

When `--random-phase-correction` is enabled, `stats["std"]` is computed from pre-correction CCC maps but used for cutoff estimation on post-correction scores. The noise correction reduces effective noise, so the pre-correction sigma is inflated, producing a cutoff that misses real particles.

Recompute variance/std from the corrected score map within the ROI after noise correction. The non-corrected path is unchanged.

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -231,6 +231,13 @@ def extract_particles(
                 f"job stat '{x}' is NaN or inf, please check your volumes and update "
                 "the job_stats json accordingly"
             )
+    if job.random_phase_correction and "sigma_noise" in job.job_stats:
+        for x in ["sigma_real", "sigma_noise"]:
+            if math.isinf(job.job_stats[x]) or math.isnan(job.job_stats[x]):
+                raise ValueError(
+                    f"job stat '{x}' is NaN or inf, please check your volumes "
+                    "and update the job_stats json accordingly"
+                )
 
     sigma = job.job_stats["std"]
     search_space = job.job_stats["search_space"]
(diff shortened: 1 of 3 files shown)


sroet:
The problem is is that we have to keep track of the variance **per sampled angle** as we are looking at maximized scores (so just doing the variance on the final result is an incorrect etimate of the variance). Your fix does not keep that into account for the the noise corrected path.

If I remember my statistics course correctly, the correct fix would be to keep track of the per angle variance on the noise scores and add it to the variance of the real score volume. So this would increase the variance, not decrease it, which also doesn't feel right. 

@McHaillet Do you remember why the noise correction does not update the variance? Was this just an oversight or a conscious decision?  (our noise substraction is also different than the [current STOPGAP implementation](https://github.com/wan-lab-vanderbilt/STOPGAP/blob/8a2d5778ae09968ce7f688993c143cf1a79fe050/src/tm/exec/stopgap_final_tm.m#L192-L206))

felipemunoz8128:
Hello @sroet. Thank you for reviewing my PR. Yes, I see now that the fix is more complicated than what I proposed originally. 

I have asked Claude to generate a document containing a derivation of the correct variance and cutoff formula for the corrected case. I only understand it up to a certain point, so I cannot fully vouch for it's validity. I include it here nevertheless in case it might be useful to you and the other developers as you consider this issue. The commit associated with this PR now contains Claude's implementation of its solution.

[logistic_cutoff_derivation.md](https://github.com/user-attachments/files/26634506/logistic_cutoff_derivation.md)

sroet:
hey @felipemunoz8128 

Thanks for sharing the document, will discuss it with the other developers. Can't promise on a timeline for this as it requires careful consideration on our statistics in general, (our CCC for example is already scaled with the std  of the tomogram in fourier space) and not all relevant developers are in the same lab anymore.

I was a bit sad to see it referencing section 9, without a section 9 being present :disappointed: 

In general for commits that were written with the help of AI, please turn on the attribution on the commits: https://code.claude.com/docs/en/settings#attribution-settings

specifically the "co-authored by ..." part. This helps me make a better informed decision on what comments/clarifications to ask when reviewing code (for example, asking about code-design/implementation detail choices is not relevant if the user is using AI, I can then just mention my own preferences  instead of needing to "prepare for a discussion"). 
It also helps me see which parts of the code development were done by AI (just the initial setup, or also any/all of the fixes I ask for, etc)

You can also do it by hand on your commit messages, by adding the 2 newlines followed by `Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>`

felipemunoz8128:
Hi @sroet. Sorry about that! Claude is now properly attributed. And about section 9 -- that was actually my fault 😅 I removed a few sections at the end that I felt were not totally necessary, but I didn't think to remove any references to them. Here's the original doc if you're interested:

[derivation_doc_full.md](https://github.com/user-attachments/files/26637021/derivation_doc_full.md)

No worries about the timeline. Looking forward to seeing what you all come up with 👍

McHaillet:
Nice discussion!

Some unclarities:
* the claude file suggests we'd need to keep track of the per angle sigma of the noise correlation 
* but looking for gumbel distribution on Wikipedia, it seems to be a distribution of the maximum, so that would imply it describes the distribution of the MIP over the angular search

I agree in general that something should be done about it though. My subtraction of the mean was a quick fix because it didn't  know how i could otherwise preserve the cutoff calculation

PR #354, opened 2026-06-04
Use warp xml CTF field (instead of OptionsCTF) to query for voltage etc.

McHaillet:
@CACNYC @rkjensen I think this should solve the problems you were having with using xmls from miss-alignment :)

### DIFF
--- a/src/pytom_tm/io.py
+++ b/src/pytom_tm/io.py
@@ -684,12 +684,12 @@ def parse_warp_xml_data(
     tilt_defocus_nodes = tree.findall(".//GridCTF/Node")
     tilt_dose_nodes = tree.findall(".//Dose")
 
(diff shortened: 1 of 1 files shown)


sroet:
Do you have reference when warp changed this xml header? And should we catch if people use the old warp?

johnskal:
For the first I don't know since I can't check right now, but for the second I don't think so.

McHaillet:
thanks to some people using miss-alignment and having issues with this, I found out that the CTF field is actually the reliable part of the warp metadata. WarpTools reads this field to construct ctf functions, so we should also use this.

McHaillet:
@sroet, nice to see you here, I assumed you hadn't returned to work yet ;)

sroet:
I haven't, but this came in at a time where I can multitask a bit

sroet (approved):
LGTM

PR #355, opened 2026-06-05
Set new conda package dir for each workflow

McHaillet:


johnskal (approved):
Sets a unique temp directory for CONDA_PKGS_DIRS and conda cache separation during test. Don’t see issues.

sroet (changes requested):
This is the wrong fix for a known conda issue: https://github.com/conda/conda-libmamba-solver/issues/925 if you want to solve this please implement the proposed workaround of that issue

McHaillet:
Thanks for checking guys, this workaround for conda solves it indeed :)

sroet (approved):
LGTM

McHaillet:
Thanks for finding this fix @sroet

PR #359, opened 2026-07-15
Whitening filter rework

McHaillet:
Main things happening here is a rework of whitening filter estimation:
- it now happens in small patches of size max(template.shape[0], 64) that are tiled over the tomogram
- this allows rejection of patches with very high variance that likely contain gold beads or other contamination and don't contribute a correct estimate of the power spectrum
- the power spectrum calculation now includes a tilt mask so that regions of the missing wedge and other 0 data areas are taken into account in calculating the power spectrum, which should give a more accurate estimate

Side things:
- 1D support for reduced_radial_grid, and bandpass, lowpass and highpass filters
- changed the interpolation in profile_to_weighting to ensure nyquist frequency is always the last interpolated values + explicitly zero the DC component

This is an example of the old and new whitening filter. Main things to note is that its a lot smoother due to the small patch size and averaging over patches. The new filter also more strongly up weights high frequencies, owing to the better estimate from the tilt mask:

whitening_filter

### DIFF
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -13,7 +13,7 @@ from pytom_tm.angles import get_angle_list
 from pytom_tm.matching import TemplateMatchingGPU
 from pytom_tm.weights import (
     create_wedge,
-    power_spectrum_profile,
+    estimate_whitening_filter,
     profile_to_weighting,
     create_gaussian_band_pass,
 )
@@ -514,19 +514,17 @@ class TMJob:
         )
         if self.whiten_spectrum and not job_loaded_for_extraction:
             logging.info("Estimating whitening filter...")
-            weights = 1 / np.sqrt(
-                power_spectrum_profile(
-                    read_mrc(self.tomogram)[
-                        self.search_origin[0] : self.search_origin[0]
-                        + self.search_size[0],
-                        self.search_origin[1] : self.search_origin[1]
-                        + self.search_size[1],
-                        self.search_origin[2] : self.search_origin[2]
-                        + self.search_size[2],
-                    ]
-                )
+            patch_size = min(64, min(self.search_size))
+            _, weights = estimate_whitening_filter(
+                read_mrc(self.tomogram)[
+                    self.search_origin[0] : self.search_origin[0] + self.search_size[0],
+                    self.search_origin[1] : self.search_origin[1] + self.search_size[1],
+                    self.search_origin[2] : self.search_origin[2] + self.search_size[2],
+                ],
+                self.ts_metadata,
+                patch_size,
+                voxel_size=self.voxel_size,
             )
-            weights /= weights.max()  # scale to 1
             np.save(self.whitening_filter, weights)
 
         # phase randomization options
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -383,6 +383,12 @@ class TMJob:
         self.template_shape = meta_data_template["shape"]
         self.mask_shape = meta_data_mask["shape"]
 
+        if len(set(self.template_shape)) != 1:
+            raise ValueError(
+                "Template is not cubic, all dimensions should be equal. "
+                f"Found template shape: {self.template_shape}."
+            )
+
         if self.template_shape != self.mask_shape:
             raise ValueError(
                 "Template and mask have a different shape in pixels. "
@@ -516,13 +522,13 @@ class TMJob:
             logging.info("Estimating whitening filter...")
             patch_size = min(max(self.template_shape[0], 64), min(self.search_size))
             _, weights = estimate_whitening_filter(
-                read_mrc(self.tomogram)[
+                tomogram=read_mrc(self.tomogram)[
                     self.search_origin[0] : self.search_origin[0] + self.search_size[0],
                     self.search_origin[1] : self.search_origin[1] + self.search_size[1],
                     self.search_origin[2] : self.search_origin[2] + self.search_size[2],
                 ],
-                self.ts_metadata,
-                patch_size,
+                ts_metadata=self.ts_metadata,
+                patch_size=patch_size,
                 voxel_size=self.voxel_size,
             )
             np.save(self.whitening_filter, weights)
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -514,7 +514,7 @@ class TMJob:
         )
         if self.whiten_spectrum and not job_loaded_for_extraction:
             logging.info("Estimating whitening filter...")
-            patch_size = min(64, min(self.search_size))
+            patch_size = min(max(self.template_shape[0], 64), min(self.search_size))
             _, weights = estimate_whitening_filter(
                 read_mrc(self.tomogram)[
                     self.search_origin[0] : self.search_origin[0] + self.search_size[0],
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -722,55 +722,52 @@ def estimate_whitening_filter(
     statistic: str = "median",
     voxel_size: float = 1.0,
 ) -> tuple[npt.NDArray[float], npt.NDArray[float]]:
-    """
-    Masked radial noise power spectrum of a tomogram, estimated on cubic patches.
+    """Estimate a whitening filter from the radially averaged noise power spectrum
+    of a tomogram, sampled on overlapping cubic patches.
 
-    Averages the power spectrum ONLY over Fourier voxels the tilt series actually
-    sampled (via the tilt mask), so the estimate is not biased by the empty
-    missing-wedge / inter-tilt-gap voxels.  Estimation is done on many overlapping,
-    windowed, mean-subtracted patches; the highest-variance patches (gold, carbon,
-    ice, volume edges) are rejected.
+    The power spectrum is averaged only over Fourier voxels the tilt series
+    actually sampled (via a tilt coverage mask), so the estimate is not biased by
+    the empty missing-wedge / inter-tilt-gap voxels. Estimation is done on many
+    overlapping, windowed, mean-subtracted patches, and the highest-variance
+    patches (e.g. gold, carbon, ice, volume edges) are rejected before averaging.
 
     Parameters
     ----------
     tomogram: npt.NDArray[float]
-        3D real-space array.
+        3D real-space array to estimate the whitening filter from
     ts_metadata: TiltSeriesMetaData
         tilt series metadata of the tomogram, used to build the coverage mask that
         restricts the radial average to Fourier voxels the tilt series actually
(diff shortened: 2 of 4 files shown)


src/pytom_tm/tmjob.py
```diff
@@ -514,19 +514,17 @@ def __init__(
...
-                    read_mrc(self.tomogram)[
-                        self.search_origin[0] : self.search_origin[0]
-                        + self.search_size[0],
-                        self.search_origin[1] : self.search_origin[1]
-                        + self.search_size[1],
-                        self.search_origin[2] : self.search_origin[2]
-                        + self.search_size[2],
-                    ]
-                )
+            patch_size = min(max(self.template_shape[0], 64), min(self.search_size))
```
sroet, line 523: any particular reason we don't check all dimensions of the template?

src/pytom_tm/weights.py
```diff
@@ -108,10 +110,12 @@ def radial_reduced_grid(
     radial_reduced_grid: npt.NDArray[float]
         fourier space frequency grid, 0 in center, 1 at nyquist
     """
-    if len(shape) not in [2, 3]:
-        raise ValueError("radial_reduced_grid() only works for 2D or 3D shapes")
+    if len(shape) not in [1, 2, 3]:
+        raise ValueError("radial_reduced_grid() only works for 1D, 2D or 3D shapes")
```
sroet: nitpick, I like to use sets for fixed `in` lookups

```suggestion
    if len(shape) not in {1, 2, 3}:
        raise ValueError("radial_reduced_grid() only works for 1D, 2D or 3D shapes")
```

src/pytom_tm/weights.py
```diff
@@ -108,10 +110,12 @@ def radial_reduced_grid(
...
     reduced_dim = shape[-1] if shape_is_reduced else shape[-1] // 2 + 1
-    if len(shape) == 3:
+    if len(shape) == 1:
+        return np.arange(0, reduced_dim, 1.0) / (reduced_dim - 1)
```
sroet: Any specific reason the `step` is a `float` in this `np.arange` call? (same question for the calls further down in this function)

(as [`numpy.arange documentation`](https://numpy.org/doc/stable/reference/generated/numpy.arange.html) says:

> When using a non-integer step, such as 0.1, it is often better to use [numpy.linspace](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html#numpy.linspace).

)

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+        can be interpolated to a different box size afterwards, so estimation
+        size and matching size are decoupled
+    overlap: float, default 0.5
+        fractional overlap between patches
+    reject_frac: float, default 0.10
+        fraction of the highest-variance patches to reject, 0 disables rejection
+    exclude: Optional[npt.NDArray[bool]], default None
+        boolean array, same shape as tomogram, True where voxels must not be used
+        (e.g. gold, carbon, contamination). Should not be used to mask out
+        biological structure, as that would bias the estimate low
```
sroet: For our default tomogram filters we use a "focus mask" for the tomogram, not an exclusion mask, should we make sure this behaves the same?

Also, this is am option that is currently not used in the codebase. Do you expect this to ever get implemented?

src/pytom_tm/tmjob.py
```diff
@@ -514,19 +514,17 @@ def __init__(
...
+            _, weights = estimate_whitening_filter(
+                read_mrc(self.tomogram)[
+                    self.search_origin[0] : self.search_origin[0] + self.search_size[0],
+                    self.search_origin[1] : self.search_origin[1] + self.search_size[1],
+                    self.search_origin[2] : self.search_origin[2] + self.search_size[2],
+                ],
+                self.ts_metadata,
+                patch_size,
```
sroet: please use keyword arguments on everything here

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+    wedge = create_wedge((length, length, length), mask_metadata, voxel_size)
+    mask = wedge > 0.05  # binarize the mask and set DC to zero
+    mask[0, 0, 0] = 0
+
+    # window (mandatory: controls leakage across the 3+ decade dynamic range)
+    win = _hann3d(length)
+    win_power = float((win**2).sum())
+
+    # collect windowed periodograms over overlapping patches
+    step = max(1, int(round(length * (1.0 - overlap))))
```
sroet: are you sure you want to use python's `round` function which can behave counter intuitive by using banker's rounding:
```python
>>> round(2.5)
2
>>> round(3.5)
4
>>> round(4.5)
4
```

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+    if len(psds) < 4:
+        raise RuntimeError(f"only {len(psds)} usable patches; reduce patch_size")
+
+    # reject high-variance patches (gold / carbon / ice / edges)
+    variances = np.asarray(variances)
+    keep = np.ones(len(psds), bool)
+    if reject_frac > 0 and len(psds) >= 10:
+        keep = variances <= np.quantile(variances, 1.0 - reject_frac)
+        if keep.sum() < 4:
+            keep = np.ones(len(psds), bool)
```
sroet: instead of doing this weird auto-turnoff of the rejection, we should probably make sure that line `801` is instead:
```
if len(psds)*(1.0-reject_frac) < 4 
```

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+
+    psd = np.mean([p for p, k in zip(psds, keep) if k], axis=0)
```
sroet: If this becomes big, you might want to instead use some way of numpy filtering instead of this list comprehension

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+    dof = int(keep.sum())
+
+    # masked radial average
+    q, prof = _masked_radial(psd, mask, length, voxel_size, statistic, dof)
+
+    def cosine_cutoff():
+        r_frac = radial_reduced_grid((length,))
+        lo, hi = 0.9, 1.0
```
sroet, line 801: it feels like these values should depend on some of the input into this function

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+
+    zs, ys, xs = (starts(n) for n in shape)
+    for z in zs:
+        for y in ys:
+            for x in xs:
+                yield (
+                    slice(z, z + length),
+                    slice(y, y + length),
+                    slice(x, x + length),
+                )
```
sroet: any reason why this has an opposite order for `xyz` than would be expected?

src/pytom_tm/weights.py
```diff
@@ -705,70 +712,186 @@ def ctf_1d(q):
...
+    prof[counts == 0] = np.nan  # empty shells -> NaN, not 0
 
-    _, power_profile = radial_average(np.abs(np.fft.rfftn(image)) ** 2)
+    if statistic == "median":
+        prof /= (1.0 - 1.0 / (9.0 * max(dof, 1))) ** 3  # median -> mean, Gamma(dof)
 
-    return power_profile
+    good = np.isfinite(prof) & (prof > 0)
+    if good.sum() >= 2:
+        prof = np.interp(q, q[good], prof[good])
```
sroet: should this error if this is not the case?

sroet (changes requested):
I think this is a nice improvement

I do have quite some comments and questions, and you're missing some coverage.

src/pytom_tm/tmjob.py
McHaillet: Yes, the template should always have identical dimensions. I thought we had a check for it, but it doesn't seem like it. I'll add that check if necesaary

src/pytom_tm/weights.py
McHaillet: Good question, I guess I wanted it to be a floating point array because it gets converted to float afterwards. But in principal we create integer numbers so maybe its better to use 1

src/pytom_tm/weights.py
McHaillet: good point, it should be a focus mask.

On the one hand if the mask is provided, it could be  nice to use it. However, the provided masks might not always be ideal for whitening filter calculation if they are very tight. I guess better to drop this path in that case.

src/pytom_tm/weights.py
McHaillet: good spot, needs to be changed

src/pytom_tm/weights.py
McHaillet: They do, it's a fraction of nyquist, it's because the top frequencies are not always accurate.

src/pytom_tm/weights.py
McHaillet: right, should be inverted. Or use d, h, w

src/pytom_tm/weights.py
McHaillet: Yes!

src/pytom_tm/tmjob.py
McHaillet: check added

src/pytom_tm/weights.py
McHaillet: Revised this section. It now does quantile estimation only if 10 psds are available and the fraction is not zero, because otherwise it's not statistically valid. After the block it now checks whether any valid psds are found, and continues if they are available. I removed the 4 boxes check.

McHaillet:
Addressed all the comments!

McHaillet:
Merged the radial_grid rework into this now!

src/pytom_tm/weights.py
```diff
@@ -697,64 +700,180 @@ def create_ctf(
...
+    q, prof = _masked_radial(psd, mask, length, voxel_size, statistic, dof)
+
+    def cosine_cutoff():
+        r_frac = radial_grid((length,))
+        lo, hi = 0.9, 1.0
+        ramp = np.clip((r_frac - lo) / (hi - lo), 0, 1)
+        return 0.5 * (1 + np.cos(np.pi * ramp))  # 1 below lo, 0 at hi
+
+    # transform into a whitening filter
+    w = np.where(prof > 0, 1 / np.sqrt(prof), np.zeros_like(prof))
```
sroet: not sure why the `zeros_like` array is initialized here

```suggestion
    w = np.where(prof > 0, 1 / np.sqrt(prof), 0.0)
```

sroet (approved):
One small comment left (assuming all changed lines are hit by tests now). Will not need another round of review

src/pytom_tm/weights.py
McHaillet: thought that was needed

PR #361, opened 2026-07-15
Add gcc installation to tutorial-tests workflow

sroet:
Added gcc installation to the workflow for tutorial tests.

no clue why the tutorials were passing before, now it also installes GCC like it does for the other tests


McHaillet (approved):
Simple, makes sense. Maybe it was included in the image before?

Issue #352, opened 2026-06-01
Read `ts autolevel` results when getting metadata from WarpTools xml


bwmr:
Hi, 

I'm really enjoying the easy parsing of CTF parameter from WarpTools in pytom-match-pick!

I was recently trying out miss-alignment, and the suggested workflow includes running `WarpTools ts_autolevel`. This changes the overall orientation the tomogram is reconstructed in, without changing the tilt angles in the xml file. Instead, it adds the parameters `LevelAngleX` and `LevelAngleY` in the file, e.g. :

```
<?xml version="1.0" encoding="utf-8"?>
<TiltSeries DataDirectory="/mnt/data/bwi/260504_Ec2168_CellTak_TEM1/warp/tomostar" AreAnglesInverted="True" PlaneNormal="-0.1681868, -0.11420458, 0.9791172" LevelAngleX="10.24746" LevelAngleY="-10.760223" Bfactor="0" Weight="1" MagnificationCorrection="1, 0, 0, 1" ImageDimensionsAngstrom="0, 0" VolumeDimensionsAngstrom="0, 0, 0" UnselectFilter="False" UnselectManual="" CTFResolutionEstimate="5.8">
```

Since the final reconstruction obtained is rotated by these angles to lie flatter in the volume, I guess they would also need to be taken into account for the per-tilt weighting scheme in this tool, when importing from the xml file. Would there be a way to implement this in `pytom-match-pick`

Many thanks for providing this excellent tool, and let me know if there's anything I can do to better describe the issue! 
Benedikt

McHaillet:
Hi Benedikt! Thanks for posting the issue here, definitely needs to be fixed. I can't guarantee a timeline, but will do my best to fix it soon

bwmr:
Hi Marten, 
thanks a lot! Let me know if there's anything you'd like me to help test, 
Benedikt

PR #362, opened 2026-07-16
Fix/tilt weighted wedge ctf centering

McHaillet:
I want to make a PR later that addresses #352, which needs to modify the rotations of CTFs to create the tilt weighted wedge. I read through that function and found it quite unreadable, so I made a few updates that hopefully improve readability.

I also noticed that the radial grid is created with its center in the middle of images/volumes, which means all the CTF and bandpass filters need to shift it back. It makes much more sense to do this at the grid creation already.

### DIFF
--- a/src/pytom_tm/weights.py
+++ b/src/pytom_tm/weights.py
@@ -83,63 +83,63 @@ def wavelength_ev2m(voltage: float) -> float:
     return _lambda
 
 
-def radial_reduced_grid(
-    shape: tuple[int, int, int] | tuple[int, int], shape_is_reduced: bool = False
+def radial_grid(
+    shape: tuple[int, int, int] | tuple[int, int],
+    reduced: bool = True,
+    fftshifted: bool = False,
+    shape_is_reduced: bool = False,
 ) -> npt.NDArray[float]:
-    """Calculates a Fourier space radial reduced grid for the given input shape, with
-    the 0 frequency in the center of the output image. Values range from 0 in the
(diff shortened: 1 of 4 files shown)


McHaillet:
Maybe easier to merge this before #359 BTW

src/pytom_tm/weights.py
```diff
@@ -83,63 +83,63 @@ def wavelength_ev2m(voltage: float) -> float:
...
+        # magnitude of frequency for a full (non-reduced) axis: 0 in the center,
+        # 1 at nyquist
+        values = np.abs(np.arange(-n // 2 + n % 2, n // 2 + n % 2, 1.0)) / (n // 2)
+        return values if fftshifted else np.fft.ifftshift(values)
+
+    def last_axis(n: int) -> npt.NDArray[float]:
+        if not reduced:
+            return full_axis(n)
+        reduced_dim = n if shape_is_reduced else n // 2 + 1
+        return np.arange(0, reduced_dim, 1.0) / (reduced_dim - 1)
```
sroet: nitpick: step should probably be an int here:

```suggestion
        return np.arange(0, reduced_dim, 1) / (reduced_dim - 1)
```

sroet (changes requested):
LGTM very nice reduction of complexity! one some comment(s)

Also, as this is changing the default behavior and API of our code, please update the version accordingly

sroet (approved):
LGTM, feel free to merge on green

PR #364, opened 2026-07-22
Fix/phase randomize template

McHaillet:
This is a PR that improves the code for generating a phase randomized template.

Most critical fix is that it ensures the power spectrum between the original and permuted template remains identical which makes it a much better decoy for correlating gold beads. This is done by generating random noise in real space and Fourier transforming it to get the new phases. 

Second fix is that it takes into account the template mask. If a mask is provided (which always happens in the current code path), the function ensures the randomization happens only in the mask.

For illustration, this is phase randomization on current `main`:

image

This is the PR without as mask:

image

And the PR with a mask:

image

### DIFF
--- a/src/pytom_tm/template.py
+++ b/src/pytom_tm/template.py
@@ -4,10 +4,7 @@ import voltools as vt
 import logging
 from scipy.ndimage import center_of_mass, zoom
 from scipy.fft import rfftn, irfftn
-from pytom_tm.weights import (
-    create_gaussian_low_pass,
-    radial_grid,
-)
+from pytom_tm.weights import create_gaussian_low_pass
 
 
 def generate_template_from_map(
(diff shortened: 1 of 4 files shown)


src/pytom_tm/template.py
```diff
@@ -118,35 +115,60 @@ def generate_template_from_map(
...
+    result = irfftn(amplitude * np.exp(1j * phase), s=t.shape)
+
+    if mask is not None:
+        for _ in range(n_iter):
+            result = result * mask
+            phase = np.angle(rfftn(result))
+            result = irfftn(amplitude * np.exp(1j * phase), s=t.shape)
+        result = result * mask
+        if result.sum() > 0:  # match total mass under the (possibly soft) mask
+            result = result * (t_eff.sum() / (result * mask).sum())
```
sroet: should we error if this is false?

sroet (changes requested):
LGTM, would like the test to also test without a mask (just making sure some of the (implicit) `else` paths are hit or make the mask non-optional if we never expect users to not use a template mask

src/pytom_tm/template.py
McHaillet: Instead of this potentially buggy check (if the template is normalized under the mask this return 0), I went for just a sign flip. We just want to ensure that the contrast of the template and phase randomized object is correct. They are later anyway normalized in template matching, so we don't care about scaling.

McHaillet:
I made the mask non-optional and the function private because it's not returning a scaled phase-randomized template. And we don't care about that since its anyway normalized.

sroet (approved):
LGTM, thanks for addressing the comments!

PR #365, opened 2026-07-27
ruff fixes

sroet:
This PR deals with the ruff update fixes that broke our tests

The one thing the requires a deeper think is the root logger, added it to the ignore for now, but will make an issue to look at that later


McHaillet (approved):
LGTM, only aestatic changes. And the root-logger will be in another fix.

I do like this new ruff style for imports, the separation between built-in, dependencies, and own.

PR #367, opened 2026-07-27
remove root logger

### LINKED ISSUE #366: fix root logger ruff failure
#365 solved most issues, but added LOG015 to the ruff ignore list.

We should really rethink our logging to be module dependent (so people know where the logs come from) but still obeys our silencing etc

sroet:
This PR:
- removes root logger
- re-enables LOG015 for ruff
- update tests to grab logs from "pytom_tm"
- added some tests for logs that missed coverage
- CLI output for running our test suite is identical between main and this pr (except for the logs comming from `pytom_tm.*` instead of `root`)

closes #366


sroet:
Log silencing for tests seems to have broken, will investigate further

McHaillet:
Let me know if I should already review this.

sroet:
@McHaillet This is now ready for a round of review

src/pytom_tm/__init__.py
```diff
@@ -3,10 +3,52 @@
...
+    logger. This should be called once by an application entry point (e.g. one of
+    the pytom_tm CLI scripts) rather than by library code.
+
+    Parameters
+    ----------
+    level: int
+        the logging level to set for the pytom_tm package logger, for example
+        logging.INFO or logging.DEBUG
+    """
+    package_logger = logging.getLogger(__name__)
```
McHaillet: line 12 already sets `logger = logging.getLogger(__name__)`. Is it needed to rerun it in the function scope?

src/pytom_tm/__init__.py
```diff
@@ -3,10 +3,52 @@
...
+    # remove any handlers that might have been added by a previous call, this
+    # mimics the `force=True` behaviour of logging.basicConfig() but scoped to
+    # the pytom_tm logger instead of the root logger
+    for handler in list(package_logger.handlers):
+        if not isinstance(handler, logging.NullHandler):
+            package_logger.removeHandler(handler)
+
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("%(levelname)s:%(name)s:%(message)s"))
+    package_logger.addHandler(handler)
```
McHaillet, line 45: The intention seems that this propagates to all the modules in the package, but does that still happen after we set propagate to False below?

McHaillet (commented):
Sweet, seems like a much needed update :)

Just had two questions about things that were not immediately clear to me.

src/pytom_tm/__init__.py
sroet: This attaches the handler to the `pytom_tm` logger (as we used to do for the `root` logger), not any loggers further down, and the `propagate` down below also **only** prevents the `pytom_tm` logger to pass messages to any handlers attached to the `root` logger

src/pytom_tm/__init__.py
sroet: You are correct, that one should suffice

src/pytom_tm/__init__.py
McHaillet: Okay good! That's what we want

McHaillet (approved):
Good to go!

McHaillet:
Once this is merged, we should tag a release!

sroet:
We probably should, do you mind doing that? Already stepped away from my computer and tagging releases is not trivial on my phone

McHaillet:
I'll do it tomorrow

PR #369, opened 2026-07-28
compatibility with warp level angles and default behaviour

### LINKED ISSUE #352: Read `ts autolevel` results when getting metadata from WarpTools xml

McHaillet:
closes #352 

This PR adds compatibility with Warp level angles and automates some default behaviour with Warp xml's.

Improved defaults:
* phase-flip CTF model is now automatically set with warp xmls because Warp tomogram are always CTF corrected
* defocus handedness is automatically set (as emperically determined, see below). The AreAnglesInverted flag in warp xmls only determines the defocus handedness. False corresponds to -1 in pytom-match-pick, True corresponds to 1. 

Compatibility with level angles required some rewrites:
* symmetric/asymmetric wedge are now combined into one function (_create_binary_wedge) that has simplified implementation. The main functionality is rotating the coordinate grids to the coordinate system of the leveled sample. Then the angle phi relative to the central plane is calculated with arctan2(x, z'). phi is used to determine what points are covered by the extreme angles of the tilt series.
* _tilt_weighted_wedge got a rewrite
   * The dose weighting is now calculated in 2D because it's unnecessary to compute it in 3D. 
   * Fixed a bug for rotation of even dimension where the center is of in Fourier space
   * Fixed a bug in full to reduced Fourier space representation, which now requires a slightly weird flip and roll afterwards. But I decided that was preferred so that we stay consistent with previous behaviour. The other option is remove that line and invert all the angles that we input to the function xD
   * Below images that show the binary wedge, tilt weigthed wedge, and warp wedge to show that we are now compatible
   * To be clear: this is not at the expense of compatibility with AreTomo/RELION5, everything is still consistent with that

Top left is our tilt weighted wedge, top right is our binary wedge, bottom left is the warp wedge. Views are xy, xz, and yz planes. Colours are different because I didn't set phase-flip=False when generating, if set to true is is the same.

 Screenshot from 2026-07-28 13-12-35

Screenshot from 2026-07-28 13-12-21

Screenshot from 2026-07-28 13-11-52

This is the defocus handedness test I did which is with AreAnglesInverted=False. From the warpylib codebase I found that switching this flag only inverts defocus handedness, so that's what I did in our codebase too:

Screenshot from 2026-07-29 11-56-07


src/pytom_tm/io.py
```diff
@@ -726,6 +739,9 @@ def _flatten(t):
         tilt_angles=flattened_tilt_angles,
         ctf_data=ctf_data,
         dose_accumulation=flattened_tilt_dose,
+        level_angle_x=level_angle_x,
+        level_angle_y=level_angle_y,
```
sroet, line 743: Do these angles flip in any way when `AreAnglesInverted` is set?

src/pytom_tm/weights.py
```diff
@@ -356,157 +359,84 @@ def create_wedge(
...
-    wedge_section[wedge_section > limit] = limit
-    wedge_section[wedge_section < -limit] = -limit
-    wedge_section = (wedge_section - wedge_section.min()) / (
-        wedge_section.max() - wedge_section.min()
-    )
-
-    # set top of the wedge
-    wedge_2d = wedge_section.copy()
+    if abs(alpha_min) > np.pi / 2 or abs(alpha_max) > np.pi / 2:
+        raise ValueError("Negative wedge angles are not defined")
```
sroet: I don't think this is a descriptive error message for this check

src/pytom_tm/weights.py
```diff
@@ -325,20 +329,19 @@ def create_wedge(
...
-        if np.round(wedge_angles[0], 2) == np.round(wedge_angles[1], 2):
-            wedge = _create_symmetric_wedge(
-                shape, wedge_angles[0], cut_off_radius
-            ).astype(np.float32)
-        else:
-            wedge = _create_asymmetric_wedge(
-                shape, (wedge_angles[0], wedge_angles[1]), cut_off_radius
-            ).astype(np.float32)
+        alpha_min = min(tilt_angles_rad) + level_angle_y_rad
+        alpha_max = max(tilt_angles_rad) + level_angle_y_rad
```
sroet, line 337: This part (including any value checking) should be done as part of the dataclass

src/pytom_tm/weights.py
```diff
@@ -356,157 +359,84 @@ def create_wedge(
...
-        2 * min(shape[0], shape[2]) // 2
-    )
-    wedge_section[wedge_section > limit] = limit
-    wedge_section[wedge_section < -limit] = -limit
-    wedge_section = (wedge_section - wedge_section.min()) / (
-        wedge_section.max() - wedge_section.min()
-    )
+    # x and y are negated to match the missing-wedge orientation established
+    # for this codebase relative to WarpTools/AreTomo reconstructions
+    x = -(np.fft.fftfreq(shape[0]) * shape[0] / (shape[0] // 2))[
```
sroet, line 412: Is the missing wedge orientation identical between Warptools and Aretomo?

tests/test_io.py
```diff
@@ -86,6 +93,50 @@ def test_correct_angle_sign(self):
...
+            'AreAnglesInverted="False"',
+            'AreAnglesInverted="False" LevelAngleX="1.5" LevelAngleY="-3.2"',
+            1,
+        )
+        with TemporaryDirectory() as tmp_dir:
+            level_angle_xml = pathlib.Path(tmp_dir).joinpath("level_angle.xml")
+            level_angle_xml.write_text(raw_xml, encoding="utf-8")
+            _, ts_metadata = parse_warp_xml_data(level_angle_xml, TEST_TOMOGRAM)
+        self.assertEqual(ts_metadata.level_angle_x, 1.5)
+        self.assertEqual(ts_metadata.level_angle_y, 3.2)
```
sroet: I would rather we test that we add a `-` instead of losing one (as that might be a parsing issue as well)

sroet (changes requested):
Awesome PR! I do have some questions, a comment on the design (value validation checks should be done inside the dataclasses), a comment on one of the tests

Also a general question:
Do we want to enable people to supply their own level_angle_x/y (in case they haven't used warp)?

src/pytom_tm/weights.py
McHaillet: Do you also mean the checks about angles not being larger than -90, +90 degrees? Because those are sort of specific to the binary wedge function

sroet: Ah true! Sorry I missed this was inside the else statement, nevermind then!

src/pytom_tm/io.py
McHaillet: They do not, that flag in the metadata just switches the defocus handedness. Not the handedness of the reconstruction.

src/pytom_tm/weights.py
McHaillet: Sadly no, that's why your PR was needed that multiplies the angles with -1. I will update this comment, because it implies WarpTools/AreTomo have identical wedge orientation.

tests/test_io.py
McHaillet: good spot, I will update it

McHaillet:
> Do we want to enable people to supply their own level_angle_x/y (in case they haven't used warp)?

Maybe further down the line. But currently, no other software implements such levelling angles. So, it could rather cause confusion. You shouldn't use the x level for a tomogram reconstructed with AreTomo because it will horribly mess up the scores.

sroet (approved):
LGTM! thanks for handling my comments

McHaillet:
Thanks for the thorough read

PR #370, opened 2026-08-04
Add gcc to release workflow

McHaillet:
Same gcc issues we had before is also causing the release to fail.


sroet (approved):
LGTM

sroet:
We would also need to bump the version number in order for the testpypi upload not to fail (We should really think about making that one more robust....)

McHaillet:
> We would also need to bump the version number in order for the testpypi upload not to fail (We should really think about making that one more robust....)

Do we? Because it didn't even get to the stage of uploading I think

sroet:
image

sroet:
give me a couple seconds before you do that, I think we should rewrite that workflow to allow for us to manually force the pypi upload (we can still remove the package on pypi if it is indeed broken and then update the version number)

sroet:
this diff should give us the option to also rerun everything except the upload to test-pypi manually if we would ever need it: 
```diff
--- a/.github/workflows/do-release.yml
+++ b/.github/workflows/do-release.yml
@@ -4,6 +4,7 @@
   release:
     types:
       - published
+  workflow_dispatch:
 
 # Following: https://docs.pypi.org/trusted-publishers/using-a-publisher/
 # and https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
@@ -35,6 +36,7 @@
   publish-to-testpypi:
     name: publish to test-pypi
     needs: build-release
+    if: github.event_name == 'release'
     runs-on: ubuntu-latest
     environment: release
     permissions:
@@ -53,6 +55,9 @@
   test-testpypi:
     name: test testpypi deployment
     needs: publish-to-testpypi
+    if: |
+      always() &&
+      (needs.publish-to-testpypi.result == 'success' || needs.publish-to-testpypi.result == 'skipped')
     runs-on: [self-hosted, linux, x64, gpu]
     container:
       image: continuumio/miniconda3
```
do you mind adding it to this PR as it affects the same file?
(diff was made with the help of Claude Sonnet 5)

sroet:
(and then manually trigger the workflow, which should skip the test-pypi upload)

McHaillet:
Yep, will add it. Thanks :)

McHaillet:
Added it, can you double check to make sure its as you intended?

sroet (approved):
Looks valid! Hope it works 🤞

PR #371, opened 2026-08-04
deal with github action "success" quirk

sroet:
addition to #370 to not skip the real release on a manually triggered workflow

background:
the default `if` is `success()` which has a quirk that if any previous step is skipped (even if it is not a direct dependency), the step itself is also skipped. [reference for the known issue](https://github.com/actions/runner/issues/2205)


McHaillet (approved):
That's a quirk for sure

PR #372, opened 2026-08-11
Add --tomogram-fanned-wedge option

### LINKED ISSUE #318: make fanned binary wedge for tomogram mask
Discussed this a bit in person yesterday. I think we agreed (but please tell me if not the case) any weighting applied to the tomogram should preferably be binary, as otherwise it interferes in Fourier space with the interpolation and weighting of the tomographic reconstruction. It is an option to apply a fanned binary wedge to the tomogram to explicitly set everything to zero where we dont expect information relative to the template weighting, and that could be nice to add in the future. Might be as easy as:

```
fanned_binary_wedge = create_wedge(tilt_angles)
fanned_binary_wedge[fanned_binary_wedge != 0] = 1
```

_Originally posted by @McHaillet in https://github.com/SBC-Utrecht/pytom-match-pick/pull/311#discussion_r2416034049_

sroet:
This PR:
- adds the `--tomogram-fanned-wedge` flag that optionally makes the tomogram filter fanned instead of a full wedge 

todo:
- [ ] add tests that uses this option and makes sure the resulting mask is similar as with a binary wedge (assuming dense angle sampling)

Question:
- currently it it possible to only apply per-tilt-weigthed to the tomo and skip it for the template, do we want to block that off?

closes #318

### DIFF
--- a/src/pytom_tm/entry_points.py
+++ b/src/pytom_tm/entry_points.py
@@ -805,6 +805,15 @@ def match_template(argv=None):
         "--relion5-tomograms-star is used.",
     )
     filter_group.add_argument(
+        "--tomogram-fanned-wedge",
+        action="store_true",
+        default=False,
+        required=False,
+        help="Use a binary version of the per-tilt-weighted wedge for the "
+        "tomogram instead of the full binary wedge defined by the minimum "
+        "and maximum tilt angles.",
+    )
+    filter_group.add_argument(
         "--voxel-size-angstrom",
         type=float,
         required=False,
@@ -1150,6 +1159,7 @@ def match_template(argv=None):
         low_pass=args.low_pass,
         high_pass=args.high_pass,
         whiten_spectrum=args.spectral_whitening,
+        tomogram_fanned_wedge=args.tomogram_fanned_wedge,
         rotational_symmetry=args.z_axis_rotational_symmetry,
         particle_diameter=args.particle_diameter,
         random_phase_correction=args.random_phase_correction,
--- a/src/pytom_tm/tmjob.py
+++ b/src/pytom_tm/tmjob.py
@@ -96,6 +96,7 @@ def load_json_to_tmjob(
         # Use 'get' for backwards compatibility
         high_pass=data.get("high_pass", None),
         whiten_spectrum=data.get("whiten_spectrum", False),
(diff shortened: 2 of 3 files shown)


sroet:
@McHaillet It seems like `_create_tilt_weighted_wedge` checks for cube-ness of the input (which isn't commonly the case for tomograms) is there anything in that function that can inherently not deal with non-cube shapes?

McHaillet:
Yes, that is correct, because the voltools transform won't correctly rotate the slice if it's non-cubic. We would need another way to do it.

src/pytom_tm/entry_points.py
```diff
@@ -804,6 +804,15 @@ def match_template(argv=None):
         "Per-tilt-weighting is always on when --warp-xml-file or "
         "--relion5-tomograms-star is used.",
     )
+    filter_group.add_argument(
```
McHaillet, line 807: I would leave this CLI option out and use the fanned wedge by default. It should be better for explicitly masking the regions where there is no expected signal if all the tilt angles are available.

sroet: I am mostly worried about a mismatch between the angles we think are sampled and the actually reconstructed tomogram angles (say by a few degrees)

McHaillet: That shouldn't be the case, unless there is a mismatch with how we assume other software makes a reconstruction. But that can quickly happen of course :|

McHaillet: Could add a path to the debug flag that writes a png of the central xz-slice of both the tomogram power spectrum and the fanned wedge

McHaillet:
I would make it that the fanned wedge is applied by default if the tilt angle information is available. But have the option to turn it off. Or just set the default to True I guess.

McHaillet (commented):
Tests and code read very clean to me. I also like the tests that check if dense tilt sampling converges to a binary min max angle mask.

Only points are:
* default to true as mentioned above, but leave the cli so that it can be turned off
* could you share an image of the fanned wegde, just for visual confirmation

sroet:
I will add the debug plotting code add the image to this PR, the default option, and a version bump (as this is a significant change of old behavior)

PR #373, opened 2026-09-09
Minor QOL improvement: automatic .tomostar suffix for Warptools

MoritzWM:
Hello,

I'm mainly using pytom to pick on Warptools reconstructed tomograms. After extraction, I always need to adjust the particles.star file and replace the `_xx.xxApx.mrc` with `.tomostar` so I can extract particles using Warptools. I wrote a wrapper for that but I still think it's a little annoying.

This PR makes pytom adjust the tomogram ID in the particles.star file if Warp xml files were passed and the file name ends in typical warp fashion.

### DIFF
--- a/src/pytom_tm/extract.py
+++ b/src/pytom_tm/extract.py
@@ -1,6 +1,7 @@
 import logging
 import math
 import pathlib
+import re
 
 import numpy as np
 import numpy.typing as npt
@@ -12,7 +13,7 @@ from scipy.special import erfcinv
 from tqdm import tqdm
 
 from pytom_tm.angles import convert_euler, get_angle_list
-from pytom_tm.dataclass import RelionTiltSeriesMetaData
+from pytom_tm.dataclass import RelionTiltSeriesMetaData, WarpTiltSeriesMetaData
 from pytom_tm.io import read_mrc
 from pytom_tm.mask import spherical_mask
 from pytom_tm.tmjob import TMJob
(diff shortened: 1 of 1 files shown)


src/pytom_tm/extract.py
```diff
@@ -331,8 +332,12 @@ def extract_particles(
 
     # remove relion5 reconstructed tomogram name as it messes with linking the tilt
     # series id when extracting subtomos
+    warp_match = re.match(r"(.*)(_\d+\.\d+Apx)", tomogram_id)
```
sroet, line 335: a small comment on what this regex is matching is nice, If I read this correctly it is matching:
`*_[1 or more numbers].[1 or more numbers]Apx` and splits out the `*` part to be used lower down`

Also please move the comment about relion5 before this line inside the `if` statement

sroet (changes requested):
Thanks for the fix, I know it is a somewhat common issue. 
Fix looks fine by me, please add a test for this behavior (so it doesn't break in the future), add a comment for the regex and move the current comment closer to the relevant code.

Let me know if you want me to write the test instead

MoritzWM:
I can write the tests, I feel like I should get some practice doing that.

MoritzWM:
I rewrote the `entrypoint test_entry_points:test_extract_candidates`. It now checks if the `rlnMicrographName` is correctly renamed only if a warp xml file is passed and the filename ends in warp fashion.

On a side note: to get  the tests running on my computer (Python 3.14 in vscode) I had to change a few things that are not in this PR. Probably because I'm running on a newer version?
1. add a `__init__.py` file to the tests directory
2. change all `testing_utils` imports to `.testing_utils`
3. change `@unittest.mock.patch` to `@path` and add `from unittest.mock import patch` at the top

sroet:
Thanks!
> On a side note: to get the tests running on my computer (Python 3.14 in vscode) I had to change a few things that are not in this PR. Probably because I'm running on a newer version?

    add a __init__.py file to the tests directory
    change all testing_utils imports to .testing_utils
    change @unittest.mock.patch to @path and add from unittest.mock import patch at the top

If I remember correctly this is needed when you run the tests from the tests directory instead of the main code directory.

sroet (approved):
LGTM, thanks for the implementation!

Let me know if you need a quick release of this code or if it can wait a bit (no release planned at the moment)

MoritzWM:
Thanks! No need to rush

PR #375, opened 2026-09-15
generate the filtered tomogram only once

McHaillet:
Colleague here was running into CPU memory issues because each GPU-worker was creating the fully filtered tomogram before launching its subvolume to the CPU. There was a comment in the code that already mentioned we can only do it once. Note that this happened specifically for a tomogram from data collected at large raw pixel size (close to 5A).


src/pytom_tm/tmjob.py
```diff
@@ -560,6 +560,10 @@ def __init__(
         # caching stuff
         self._tomogram_filter = None
         self._template_filter = None
+        # path to a fully filtered (bandpass + whitening + wedge) copy of the
+        # tomogram, shared by all sub jobs so the filtering only has to happen once,
+        # see _generate_filtered_tomogram()
+        self.filtered_tomogram = None
```
sroet: style wise, this should probably start with an underscore as we never want the user to interact with this

src/pytom_tm/tmjob.py
```diff
@@ -905,12 +956,12 @@ def start_job(
...
-            tuple([next_fast_len(s, real=True) for s in tomo.shape]),
-            dtype=np.float32,
-        )
-        fast_tomo[: tomo.shape[0], : tomo.shape[1], : tomo.shape[2]] = tomo
+        if self.filtered_tomogram is not None:
+            # tomogram was already fully filtered once for all sub jobs sharing
+            # this search, see TMJob._generate_filtered_tomogram()
+            fast_tomo = read_mrc(self.filtered_tomogram)
+        else:
+            fast_tomo = self._filter_full_tomogram()
```
sroet: I think this can be nicer solved by using a `property` on the class

tests/test_tmjob.py
```diff
@@ -475,6 +476,42 @@ def test_filter_caching(self):
         with NamedTemporaryFile(suffix=".json") as temp:
             job.write_to_json(pathlib.Path(temp.name))
 
+    def test_filtered_tomogram_generated_once(self):
```
sroet, line 479: I really like the `patch` use in this test 😄

sroet (changes requested):
In general I think `tmjob.filtered_tomogram` should start with an underscore everywhere to indicate users shouldn't touch it.
I also mentioned maybe using a `property` for `tmjob.filtered_tomogram` (no underscore) to do the `if None: generate` logic.

sroet:
Other than those comments, great addition to the code!

McHaillet:
I think this is what you had in mind with the property, but correct me if Im wrong

sroet (approved):
LGTM

